PR #14, opened 2026-05-06
fix: record model on streaming Usage so runs.model populates

denn-gubsky:
## Summary

The streaming \`message_delta\` handler builds the final cumulative \`providers.Usage\` from the wire's usage object — but Anthropic's streaming usage doesn't carry the model identifier. So every \`runs.model\` row in the DB has been empty, even though \`req.Model\` is sent on every call.

This PR plumbs \`req.Model\` from \`driver.Call\` → \`streamEvents\` → \`processFrame\` and stamps it onto Usage in the message_delta case. Three small changes; one new test pinning the invariant.

## Why this matters

Cost incident 2026-05-06: when investigating which models were called during specific time windows, every loomcycle run reported \`model = ""\`. Post-incident accounting had to guess from \`session.agent\` + \`sample.yaml\`'s per-agent model assignment. No ground truth.

After this PR, queries like:

\`\`\`sql
SELECT model, COUNT(*), SUM(input_tokens), SUM(output_tokens)
FROM runs
WHERE started_at > strftime('%s','now','-1 day') * 1000000000
GROUP BY model;
\`\`\`

return real per-model breakdowns. Combined with the price table in jobs-search-web's \`src/lib/pricing.ts\`, this gives ground-truth cost-per-model for retrospectives.

## Test plan

- [x] \`go test -count=1 ./...\` — all 18 packages pass
- [x] New \`TestStreamingUsage_PopulatesModel\` asserts \`ev.Usage.Model\` equals \`req.Model\` on EventDone
- [x] Reverting the plumbing makes the test fail with \`Usage.Model = ""\` vs the requested model

### DIFF
--- a/internal/providers/anthropic/driver.go
+++ b/internal/providers/anthropic/driver.go
@@ -104,7 +104,15 @@ func (d *Driver) Call(ctx context.Context, req providers.Request) (<-chan provid
 	}
 
 	out := make(chan providers.Event, 16)
-	go streamEvents(ctx, resp.Body, out)
+	// Thread req.Model into the stream parser so the message_delta
+	// handler can stamp it onto the Usage event. Anthropic's
+	// streaming usage object doesn't carry the model field; without
+	// this plumbing, every providers.Usage on the wire has Model="",
+	// the loop's totalUsage.Model stays empty, and the runs table's
+	// `model` column never populates. Cost incident 2026-05-06: with
+	// no model recorded per run, we couldn't verify which agent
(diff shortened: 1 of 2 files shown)


denn-gubsky:
Superseded by 5bdccfc ("fix(anthropic): plumb model alias from message_start into final Usage") which landed on main directly. Same root cause; the main commit captures the wire alias from the message_start event rather than echoing req.Model, which matches what Anthropic actually billed against. Closing + deleting the branch.

PR #49, opened 2026-05-10
feat(config): discover agents from a directory of MD files

denn-gubsky:
## Summary

Eliminate the synchronisation pain between yaml `agents:` entries and Claude Code `.md` files. Loomcycle now scans a directory of `<name>.md` files (`LOOMCYCLE_AGENTS_ROOT`), parses each frontmatter as the base AgentDef, and uses the body as the system prompt. The yaml `agents:` map remains an optional override layer for environment-specific tweaks.

Single source of truth in normal operation; per-deployment overrides when needed.

## Why

Two-source-of-truth has been recurring operational pain:

- 2026-05-09 VM bring-up: 4 agent files existed in dev branch but not main → loomcycle restart loop until the workspace was switched to dev.
- Broader pattern: yaml entries pointing at `system_prompt_file` paths that don't exist, or MDs without matching yaml entries, both fail in different ways.

The frontmatter and yaml are the same shape of information, asked twice. This PR collapses it.

## Design (user-confirmed)

| Question | Answer |
|---|---|
| Yaml `agents:` role | **Override layer.** MDs are the base; yaml entries override per-agent fields when present. All MDs are enabled by default. Mixed-mode and yaml-only deployments stay supported. |
| Frontmatter shape | **Flat top-level keys.** Loomcycle-specific fields (`tier`, `models`, `skills`, `max_tokens`, `memory_scopes`, …) live alongside Claude Code's standard keys (`name`, `description`, `tools`, `model`). Claude Code ignores unknown frontmatter keys, so MDs stay portable across both consumers. |
| Tool list shape | Both `tools: A, B, C` (Claude Code's comma-string) AND `allowed_tools: [A, B, C]` (loomcycle's yaml list) are accepted. When both present, `allowed_tools` wins. |
| Discovery | All `<name>.md` files in the directory are enabled. To disable one, move the file out. |

### New surface

- **`internal/agents/loader.go`** — `LoadSet(root)`, `Set.Get`, `Set.Names`, `ParseToolList(s)`. Mirrors `internal/skills/loader.go` shape: same delimiter rules, same body-only fallback, same name-validation reject set.
- **`Env.AgentsRoot`** in `internal/config/config.go` — sourced from `LOOMCYCLE_AGENTS_ROOT`.
- **`config.discoverAgents(cfg, root)`** — walks the directory, converts each `agents.Agent` to `config.AgentDef`, merges with yaml entries.
- **`config.mergeAgentDef(base, override)`** — per-field shallow merge with yaml-overrides-discovered semantics. `nil` slice = absent (keep discovered); non-nil = explicit override (replace). Special case: yaml `system_prompt_file` clears discovered `system_prompt` (body) so `resolveSystemPromptFiles`' "both set" check doesn't fire.

### Resolution order

```
config.Load(path)
├─ defaults populated
├─ if path != "":
│   ├─ os.ReadFile + expandEnv + yaml.Unmarshal
│   ├─ stash configDir
│   ├─ NEW: read LOOMCYCLE_AGENTS_ROOT inline
│   │   if set: discoverAgents(cfg, root)
│   │     └─ agents.LoadSet(root)
│   │     └─ for each: cfg.Agents = mergeAgentDef(discovered, yaml)
│   └─ resolveSystemPromptFiles (sees merged map)
├─ Env struct loading (fills cfg.Env.AgentsRoot for observability)
├─ resolveSkills (sees merged map)
├─ validate (covers all merged agents — Pin XOR Tier rule catches MD/yaml conflicts)
└─ return
```

The env var is read inline because the env-block populates later in Load(); restructuring that order would risk subtler regressions for one early reader.

### Operator changes

- New env var: `LOOMCYCLE_AGENTS_ROOT=/srv/loomcycle/agents`.
- New boot log line: `agents: discovered from <root> — total in cfg.Agents (after yaml merge): N`.
- `.env.example` documents the canonical frontmatter key list (mitigation for the typo'd-key silent-zero risk).

## Out of scope (deferred)

- No yaml↔MD bidirectional sync; no MD generation tool; no `loomcycle agents lint` subcommand; no SIGHUP hot-reload.
- No nested-directory layout — flat `<name>.md` only, matches Claude Code convention.
- No strict-frontmatter enforcement (`yaml.KnownFields(true)`); typo'd keys silently zero today. Follow-up PR would also tighten the skills loader.
- No parallel jobs-search-agent migration — per CLAUDE.md cross-repo discipline, lands in loomcycle first; jobs-search-agent migration follows on a separate branch.

## Test plan

- [x] `internal/agents/loader_test.go` — 13 unit tests (happy path, missing/nonexistent root, malformed frontmatter, no-closing-delim, body-only, comma-string tools, `allowed_tools` wins, filename/name mismatch, subdir + non-md skip, empty body, YAML-list tools, tools rejects bad type).
- [x] `internal/config/config_test.go` — 6 cross-cutting tests (discovery + yaml merge, discovery only, yaml-only regression guard, Pin+Tier validation post-merge, `system_prompt_file` overrides discovered body, discovered skills bundle correctly).
- [x] `go test ./...` — all 25 packages green.
- [x] `go test -race ./internal/agents/ ./internal/config/` — clean.
- [x] Manual: `loomcycle validate <yaml>` with `LOOMCYCLE_AGENTS_ROOT` set → discovered agent appears in summary. Boot log emits the new line.
- [ ] Manual end-to-end: real agent run via the discovered agent (deferred to next session, paired with the jobs-search-agent migration).

## Risks

- **Typo'd frontmatter key silently drops to zero.** A misspelled `provdier:` parses as `Provider=""`, looks like "operator forgot to set it". Mitigation: documented key list in `.env.example`. Strict mode is a follow-up PR (would also affect skills).
- **AGENTS_ROOT set but empty dir** → empty Set, no error. If yaml also has zero `agents:`, validate() may pass with zero agents → API-time "unknown agent" error when a request arrives. Recommendation: keep silent-empty (consistent with `SkillsRoot=""`); the runtime error is loud enough.
- **Filename collisions across cases** (`foo.md` vs `Foo.md`) — distinct on Linux, same on macOS/Windows. Loader treats them as distinct names; same caveat exists for skills today.
- **Body containing `---` separators** — parser uses first `\n---\n` after the opening. An MD with nested frontmatter triggers early body cutoff. Document but don't fix; matches skills behaviour.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -83,6 +83,44 @@ LOOMCYCLE_DATA_DIR=./data
 # not point this at a directory writable by untrusted users.
 # LOOMCYCLE_SKILLS_ROOT=/srv/loomcycle/skills
 
+# ─── Agent directory discovery (v0.8.x+) ─────────────────────────────
+#
+# Directory of flat `<name>.md` files. When set, loomcycle parses each
+# file's YAML frontmatter as the base AgentDef and the body (after the
+# closing `---`) as the system prompt. The yaml `agents:` map remains
+# OPTIONAL and acts as an override layer — yaml entries with the same
+# name override per-field. Operators get a single source of truth in
+# the MD with targeted per-environment tweaks in yaml when needed.
+#
+# Why: synchronising the yaml `agents:` block with the MDs referenced
+# from `system_prompt_file` is recurring operational pain (every
+# dev↔main branch divergence breaks loomcycle on the deploy box).
+#
+# Frontmatter contract — these top-level keys are read by loomcycle
+# (Claude Code ignores unknown keys, so the same MD drives both):
+#   name                  # string; must equal filename sans .md
+#   description           # string; informational
+#   tools                 # string "A, B, C" OR list [A, B, C]
+#   allowed_tools         # list — wins when both `tools` and this set
+#   provider              # string; e.g. anthropic / openai / deepseek
+#   model                 # string; alias or full model ID
+#   tier                  # one of low / middle / high
+#   models                # map<tier-name, [{provider, model}, ...]>
+#   effort                # one of low / medium / high
+#   max_tokens            # int; per-iteration output cap
+#   skills                # list of skill names (LOOMCYCLE_SKILLS_ROOT)
+#   memory_scopes         # list — subset of {agent, user}
+#   memory_quota_bytes    # int; per-agent override
+#   providers             # list — per-agent ProviderPriority override
+#   system_prompt_file    # string; alternative to body
+#
+# Validation rules (Pin XOR Tier, Effort domain, MemoryScopes domain,
+# etc.) run unchanged on the merged map. A typo'd key is silently
+# dropped to zero today — keep that in mind when an agent doesn't
+# behave as expected.
+#
+# LOOMCYCLE_AGENTS_ROOT=/srv/loomcycle/agents
+
 # ─── Memory tool (v0.8.0+) ───────────────────────────────────────────
 #
 # The Memory tool exposes persistent agent-scoped key/value storage to
--- a/cmd/loomcycle/main.go
+++ b/cmd/loomcycle/main.go
@@ -156,6 +156,14 @@ func main() {
 	if cfg.Env.SkillsRoot != "" {
 		log.Printf("skills: loaded %d from %s", len(skillSet.Names()), cfg.Env.SkillsRoot)
 	}
+	if cfg.Env.AgentsRoot != "" {
+		// Agents-from-MD discovery happened inside config.Load (must run
+		// before resolveSystemPromptFiles so the merged map flows through
+		// the existing pipeline). Log the count post-merge so operators
+		// see the final cfg.Agents size, which may include yaml-only
+		// entries on top of the discovered ones.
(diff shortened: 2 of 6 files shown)


denn-gubsky:
**Code review fixes pushed in `002ac42`** addressing all 4 findings from the review:

| # | Severity | What | Fix |
|---|---|---|---|
| 1 | Critical | `discoverAgents` skipped when `Load("")` called (env-only mode) | Pull discovery + `resolveSystemPromptFiles` out of `if path != "" {}` block |
| 3 | Important | `mergeAgentDef` asymmetric: yaml inline `system_prompt` didn't clear discovered `system_prompt_file` | Each prompt-source override now clears the OTHER source symmetrically |
| 2 | Should-fix | Test gap for `tools: [{a:b}, "Read"]` map-element typo | New `TestParseAgent_ToolsRejectsMapElement` |
| 5 | Should-fix | Test gap for empty-yaml-list-clears-discovered nil/empty distinction | New `TestDiscoverAgents_EmptyYAMLListClearsDiscovered` (also pins gopkg.in/yaml.v3's nil/non-nil-empty contract for future upgrades) |

Also added `TestDiscoverAgents_NoYAMLPath` as the explicit regression guard for fix #1 (the headline "MDs as sole source of truth" deployment shape).

All 25 packages green; `-race` clean on `internal/agents/` + `internal/config/`.

PR #128, opened 2026-05-17
feat(hooks): per-call host-widening capability for Pre-hooks (v0.8.17)

denn-gubsky:
## Summary

Pre-hooks can now grant the model permission to fetch hostnames outside the operator's static allowlist and the caller's `runRequest.allowed_hosts`, **per tool call**, gated on explicit operator-yaml opt-in. Solves the dynamic-discovery problem for `job-searcher` / `company-explorer` agents that find URLs in search results that nobody pre-enumerated.

## Why

`jobs-search-agent` integrators kept hitting the host-allowlist wall: the operator's static list can't enumerate every job-board / company-site / ATS-tracker domain in advance, and the caller's per-run allowed_hosts only knows what was known at run-create time. The agent finds a new URL, tries to fetch, and gets blocked. Existing options (pre-enumerate everything; widen the static list; disable host policy entirely) either don't scale or degrade the security posture too far.

The trust-boundary invariant (CLAUDE.md rule #8) prevents the model from unilaterally widening. The clean solution: delegate the per-URL allow/deny decision to a **trusted out-of-band party** — the calling service's own policy webhook — and let the OPERATOR opt that webhook in via yaml. The runtime caller and the model both cannot enable widening; only operator yaml can.

## What changed

Eight commits, broken down for review:

1. **`feat(hooks/types): add AllowHosts to PreHookResult (wire shape only)`** — extend `PreHookResult` with optional `allow_hosts []string`. Doc-comment captures the load-bearing invariants (per-call scope, leading-dot suffix semantics, confused-deputy warning). No behavior change yet.
2. **`feat(hooks): operator opt-in for host-widening (PermitHostWiden.Owners)`** — new `hooks.permit_host_widen.owners` yaml block + `LOOMCYCLE_HOOKS_PERMIT_HOST_WIDEN_OWNERS` env. `Registry.IsHostWidenPermitted(owner)` exposes the set; default-deny stance preserved for back-compat.
3. **`feat(hooks): dispatcher accumulates allow_hosts from permitted owners`** — `RunPre` unions per-call grants from permitted hooks, drops un-permitted ones with a WARN log + counter, deny discards all prior grants. `Dispatcher.Stats()` exposes `HostWidenPermitted` / `HostWidenDenied` atomic counters.
4. **`feat(tools): WithExtraAllowedHosts ctx helper + loop wires it pre-Execute`** — ctx-scoped extras attached by `dispatchOneTool` JUST before calling Execute. Sub-agents and the next iteration see the original ctx without extras (per-call scope).
5. **`feat(httptool): consult ctx-extras at host-allowed gate + redirect`** — primary gate + CheckRedirect both consult `tools.ExtraAllowedHosts(ctx)`. New `hostAllowedExtras` matcher: dotless = exact-match, leading-dot = suffix-match (intentionally stricter than the operator allowlist).
6. **`feat(loop+providers): EventHostWidened audit event`** — typed `host_widened` event with the full payload (tool_call_id, tool_name, URL, hook_owner, hook_name, hosts_added). Persisted via `makeRecordingEmit`.
7. **`docs: per-call host widening capability + sample yaml`** — full `docs/TOOLS.md` subsection covering wire shape, semantics, audit event, counters, the four known limitations, and the confused-deputy security warning. Sample yaml shows the operator opt-in shape.
8. **`chore: gofmt struct-literal alignment in hooks package`** — purely cosmetic.

## Wire shape

```json
// Pre-hook response (new optional field):
{
  "input": {...},
  "deny": {...},
  "allow_hosts": ["acme.com", ".trusted-cdn.com"]
}
```

Matching: bare = exact-hostname, leading-dot = suffix-match.

## Operator opt-in

```yaml
hooks:
  permit_host_widen:
    owners: ["jobs-search-web"]
```

Or env: `LOOMCYCLE_HOOKS_PERMIT_HOST_WIDEN_OWNERS=jobs-search-web,company-research` (APPENDS to yaml).

## Trust-boundary preservation

- Operator yaml is the SOLE enabler. Model and runtime caller cannot configure widening.
- Hook author warning lives in the `PreHookResult.AllowHosts` doc-comment: "MUST NOT be derived blindly from `tool_call.input` — the URL is model-generated and untrusted. Validate independently."
- Audit event includes the requesting URL so operators detect confused-deputy patterns post-hoc.
- Aggregate counter (`hooks_host_widen_denied`) makes un-authorised widening attempts graphable without scraping events.
- Dial-time private-IP block is NOT bypassable (a hook approving `localhost` still fails at dial unless `HTTPPrivateHostAllowlist` separately opts in).
- Sub-agents do NOT inherit a parent's widening. Per-call scope, no propagation.

## Known limitations (v1)

1. **Redirect to a brand-new host not approved up-front fails.** The hook is invoked once per tool call, before Execute; redirects within the call consult the SAME ctx-extras. Documented in `docs/TOOLS.md`.
2. **No server-side cache.** Every fetch to an unknown host fires a fresh callback. Hooks should cache client-side.
3. **No sub-agent inheritance.** Per-call scope means the grant evaporates the moment Execute returns. Sub-agents needing the same widening register their own hook against the sub-agent's name.
4. **Dial-time IP block remains orthogonal.** Operator must separately opt in via `HTTPPrivateHostAllowlist` for hosts resolving to private IPs.

## Test plan

- [x] `go build ./...` clean.
- [x] `go vet ./...` clean.
- [x] `gofmt -l .` empty.
- [x] `go test -race -timeout 120s ./...` green on the full tree.
- [x] **15 new unit + integration tests** across `internal/hooks/`, `internal/tools/`, `internal/tools/builtin/`, `internal/loop/`. Cover: wire shape round-trip, registry permit semantics, dispatcher accumulation (permitted/un-permitted/deny+allow/union/mixed/case-norm), ctx helper round-trip + scope isolation, httptool enforcement (allows / empty-no-widen / IP-block-unbypassable / redirect-allows / redirect-unknown-blocked), `hostAllowedExtras` matcher table (12 cases), end-to-end loop integration (permitted approves; un-permitted has no effect).
- [x] Sample yaml loads cleanly: `./bin/loomcycle --config loomcycle.example.yaml --version` boots.

## File touches

```
 docs/TOOLS.md                           |  91 ++++++++++++-
 internal/api/http/server.go             |   7 +-
 internal/config/config.go               |  53 ++++++++
 internal/hooks/dispatcher.go            | 148 ++++++++++++++++++-
 internal/hooks/dispatcher_test.go       | 218 ++++++++++++++++++++++++++++++++
 internal/hooks/registry.go              |  52 +++++++-
 internal/hooks/registry_test.go         |  63 +++++++++
 internal/hooks/types.go                 |  39 +++++-
 internal/hooks/types_test.go            | 113 +++++++++++++++++
 internal/loop/hooks_test.go             | 194 ++++++++++++++++++++++++++++
 internal/loop/loop.go                   |  55 +++++++-
 internal/providers/provider.go          |  64 ++++++++++
 internal/tools/builtin/httptool.go      |  65 +++++++++-
 internal/tools/builtin/httptool_test.go | 199 +++++++++++++++++++++++++++++
 internal/tools/tool.go                  |  36 ++++++
 internal/tools/tool_test.go             |  62 +++++++++
 loomcycle.example.yaml                  |  28 ++++
 17 files changed, ~1480 insertions, ~21 deletions
```

### DIFF
--- a/docs/TOOLS.md
+++ b/docs/TOOLS.md
@@ -584,11 +584,98 @@ DELETE /v1/hooks/{id}          // remove a registration
 
 ### Trust-boundary invariants
 
-- Hooks run **after** the policy layer (`allowed_tools` / `allowed_hosts`). Hooks may narrow further or rewrite content — **never widen**.
+- Hooks run **after** the policy layer (`allowed_tools` / `allowed_hosts`). Hooks may narrow further or rewrite content. The **only** exception is the per-call host-widening capability below, which requires explicit operator opt-in.
 - Hooks **cannot** tear down the agent run. The worst they can do is short-circuit one tool call with a synthetic `IsError: true` result.
 - Webhook payloads include `agent_id` and `user_id` for correlation but **do NOT** include the agent's prompt or message history.
 
-References: `internal/hooks/`, `internal/api/http/hooks.go` (registration routes), `internal/loop/loop.go::dispatchOneTool` (chain invocation).
+### Per-call host widening (v0.8.17)
+
+A Pre-hook can grant the model permission to fetch a hostname that's *not* on the operator's static `LOOMCYCLE_HTTP_HOST_ALLOWLIST` and *not* on the runtime caller's `runRequest.allowed_hosts`. The grant is **per-tool-call** — scoped to exactly one `Execute()`, never cached server-side, never inherited by sub-agents.
+
+This solves the dynamic-discovery problem: `job-searcher` finds a job posting on a site nobody pre-enumerated; the calling service's hook checks its own per-user allowlist + reputation oracle + business rules and approves the URL on the spot.
+
+#### Opt-in (operator yaml only)
+
+```yaml
+hooks:
+  permit_host_widen:
+    owners: ["jobs-search-web"]
+```
+
+Or env: `LOOMCYCLE_HOOKS_PERMIT_HOST_WIDEN_OWNERS=jobs-search-web,company-research`
+
+- **Exact-string match** against `Hook.Owner`. No globs.
+- Without an entry here, ANY hook's `allow_hosts` field is silently dropped at the dispatcher (with a `hooks: pre ... dropping grant` WARN log + counter increment via `Dispatcher.Stats().HostWidenDenied`). The model and the runtime caller cannot enable this — only the operator yaml can.
+- **Caller-authoritative mode interaction**: when `LOOMCYCLE_HTTP_CALLER_AUTHORITATIVE=true`, a permitted hook STILL widens on top of the caller's authoritative list. Operator yaml is the floor for *behaviors* (including hook delegation), not just for the static host list.
+
+#### Wire shape
+
+A Pre-hook returns an optional `allow_hosts` field alongside the existing `input` / `deny`:
+
+```json
+{
+  "input": {...},
+  "deny": {"is_error": true, "text": "..."},
+  "allow_hosts": ["acme.com", ".trusted-cdn.com"]
+}
+```
+
+#### Matching semantics
+
+Intentionally **stricter** than the operator allowlist:
+
+- `"acme.com"` (no leading dot) → **exact hostname match only**. Won't widen `careers.acme.com`.
+- `".acme.com"` (leading dot) → **suffix-match** (matches `acme.com` AND any subdomain). Symmetric with the operator-list shape.
+
+This lets cautious hook authors be surgical ("approve only the literal URL the model asked about") while supporting the broader case via the leading-dot opt-in.
+
+#### Composition
+
+- Multiple permitted-owner hooks contributing `allow_hosts` in one chain → the dispatcher returns the **deduplicated UNION** (case-insensitive on hostname).
+- A `deny` anywhere in the chain **discards all prior** `allow_hosts` grants (we don't carry policy widenings into a denied call — confused-deputy hardening).
+- Order of `AllowHosts` in the outcome is first-seen across the Pre chain.
+
+#### Audit event
+
+The loop emits one `EventHostWidened` per dispatched call where widening fired:
+
+```json
+{
+  "type": "host_widened",
+  "host_widening": {
+    "tool_call_id": "lc-0-0",
+    "tool_name": "WebFetch",
+    "url": "https://acme.com/careers/123",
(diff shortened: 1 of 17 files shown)


denn-gubsky:
Code review pass complete. Two confidence ≥ 80 findings, both fixed in commit 1ed0b18:

**1. Stale single-writer comment in `executePendingTools`** — the comment in `loop.go:834-837` said all `emit()` calls happen from the main goroutine. That was true pre-v0.8.4; v0.8.4 broke the invariant (Channel tool emits from `Execute()` goroutines) and this PR's `EventHostWidened` extends the pattern. Production is safe (makeRecordingEmit has a mutex), but the stale comment would have misled future test authors. Comment rewritten to name both v0.8.4 + v0.8.17 precedents and warn that test collectors with `ToolParallelism > 1` need their own mutex.

**2. Missing FailClosed-timeout test for AllowHosts discard** — `TestDispatcher_PreAllowHosts_DenyDiscardsPriorGrants` covered the explicit-deny path; the FailClosed-induced-deny path was not pinned by a test. The behaviour is correct today (FailClosed early-returns from `RunPre` before the AllowHosts accumulation), but a future refactor "preserving AllowHosts across FailClosed for observability" would silently widen policy. New test `TestDispatcher_PreAllowHosts_FailClosedTimeoutDiscardsPriorGrants` locks the property.

**No issues found on:**
- Trust-boundary injection — the model and runtime caller have no path into `ExtraAllowedHosts`.
- Redirect ctx capture — `CheckRedirect` correctly closes over `execCtx` (the derived ctx with extras).
- Race conditions in `dispatchOneTool` — each goroutine has its own derived ctx; counters are atomic.
- Empty/nil edge cases — `WithExtraAllowedHosts(nil)` is a no-op; `hostAllowedExtras` returns false on nil; `normaliseHost` drops empty/whitespace entries.
- Wire-shape stability — `allow_hosts` is `omitempty`; existing hooks see no wire change.
- Audit event correctness — `extractToolURL` empty-URL fallback is intentional and documented.
- Docs accuracy — all four "known limitations" match the implementation exactly.

Full `gofmt -l .` empty, `go vet ./...` clean, `go test -race -timeout 120s ./...` green on this branch.

PR #188, opened 2026-05-22
feat(lookup): canonical resolver — consolidate static-vs-dynamic load paths + prevent future drift

denn-gubsky:
## Summary

Three recent PRs patched the same architectural drift one symptom at a time:

| PR | Symptom |
|----|---------|
| #184 | `/v1/runs` returned "unknown agent" for substrate-registered names. Two bugs in one: (a) `lookupAgent` missed the substrate tier; (b) `json.Unmarshal(row.Definition, &configAgentDef)` silently dropped every field (`config.AgentDef` carries yaml-only tags; substrate persists snake_case json). |
| #186 | Substrate-registered agents lost their system prompt at every skill-enabled run. `resolveSkillBodiesForRun` rebuilds `SystemPrompt = SystemPromptBase + skill bodies`, and the dynamic load path left `SystemPromptBase` empty. |

Root cause: loomcycle's static-yaml era had ONE canonical load path:

```
yaml → config.LoadConfig → resolveSkills / resolveAgent → cfg.Agents
```

v0.8.15 added `dynamic_agents` and v0.8.22 added the AgentDef substrate. Both new READ paths SKIPPED the normalizer chain that `resolveSkills` applied at boot. Future fields added to `config.AgentDef` requiring static-path normalization would silently break for the same class of bug.

This PR consolidates the lookup chain + normalizer chain + JSON adapter into a new `internal/lookup` package, applies the SystemPromptBase fix at the write side too (so new rows persist correctly + a boot-time backfill heals legacy rows), and adds equivalence + drift tests that catch future divergence at PR-time rather than in production.

## Architecture

The 4-pillar fix (one commit per pillar):

1. **`internal/lookup` package** — `lookup.Agent` / `lookup.Skill` / `lookup.MCPServer` are the canonical name → effective-def resolvers. Each walks the static/dynamic precedence chain and applies the normalizer chain. `lookupAgent` in `server.go` is now a 5-line wrapper.
2. **Substrate adapters co-located** — `lookup.SubstrateAgentDef` / `SubstrateSkillDef` / `SubstrateMCPServer` live alongside the resolvers. Used wherever JSON unmarshal of substrate rows happens; the `skillDefOverlay` duplicate in `server.go` collapsed to use `lookup.SubstrateSkillDef`.
3. **Write-side persistence of `SystemPromptBase`** — `mergedDef.normalize()` in `agentdef.go` fills the field at every write site (execCreate / execFork / bootstrapStatic). Boot-time `BackfillAgentDefSystemPromptBase` heals legacy rows. The read-side `lookup.NormalizeAgentDef` (extracted from PR #186) stays as defense-in-depth.
4. **Equivalence + drift tests** — `TestAgent_EquivalenceYamlVsSubstrate` catches the entire class of bugs. `TestAgent_DriftDetection` (reflection-based) prevents future field-addition gaps by failing CI when `SubstrateAgentDef` and the expected json-tag set diverge. Equivalent tests for `SubstrateSkillDef` + `SubstrateMCPServer`.

## What's grouped from prior PRs

| Prior PR | What it did | What this PR does to it |
|----------|-------------|--------------------------|
| #184 lookup fallthrough | Added substrate tier to `lookupAgent` + `substrateAgentDef` adapter. | Moved to `internal/lookup`. Now the canonical home; future substrates follow the same pattern. |
| #186 SystemPromptBase read-fix | Added `normalizeSystemPromptBase` to the dynamic read path. | Moved to `lookup.NormalizeAgentDef`. Joined by a write-side `normalize()` + a boot-time backfill so the on-disk data also converges. |
| #185 skill error UX | Improved error message when static skill set is empty but substrate has names. | Unchanged — different concern (operator-facing UX, not data path). Stays as-is. |

## Audit of OTHER dynamic substrates

The Explore agent inventoried `SkillDef` and `MCPServerDef` for the same patterns. Findings:

| Substrate | Tag mismatch? | Boot normalizer gap? | Lookup tier gap? |
|-----------|---------------|----------------------|-------------------|
| AgentDef  | **Was real (PR #184)** | **Was real (PR #186)** | **Was real (PR #184)** |
| SkillDef  | Symmetric (both marshal + unmarshal go through json-tagged structs); WONTFIRE | resolveSkillBodiesForRun is the canonical runtime path; no static-only normalizer to leak | Dual-aware already in resolveSkillBodiesForRun |
| MCPServerDef | Symmetric; WONTFIRE | No boot-time normalizations needed (no SystemPromptBase-style invariant) | Pool's build callback is dual-aware |

The drift tests are forward-looking insurance: a future refactor that introduces an asymmetric intermediate struct for either of the two currently-symmetric substrates would fail the drift test before shipping.

## Verification

End-to-end smoke against a built binary (cleaned up after):

```
$ sqlite3 data/loomcycle.db "INSERT INTO agent_defs ... \
    definition='{\"system_prompt\":\"be helpful\",...}'"  # legacy row, no system_prompt_base
$ ./bin/loomcycle --config ...
2026/05/22 18:09:07 agent_defs: backfilled 1 rows with system_prompt_base
$ sqlite3 data/loomcycle.db "SELECT definition FROM agent_defs ..."
{"allowed_tools":[...],"system_prompt":"be helpful","system_prompt_base":"be helpful"}
```

Tests:
- `go test ./internal/lookup/ -v` — 6 cases: equivalence, legacy-row-fill-on-read, AgentDef drift, SkillDef drift, MCPServerDef drift, nested SubstrateMCPServerTool drift.
- `go test ./internal/store/sqlite/ -run 'BackfillAgentDefSystemPromptBase' -v` — contract test pinning the 3-case backfill semantics + idempotency.
- `go test ./internal/api/http/ -run 'TestLookupAgent' -v` — all 5 pre-existing PR #184/#186 regression tests still pass against the extracted resolver.
- `go test ./... -timeout 180s` — full pass.
- `gofmt -l .` — clean.

## Notes for review

- **No yaml/json tag unification on `config.AgentDef`** — tempting to add json tags everywhere, but every external consumer (Python adapter, gRPC, TS adapter) depends on current snake_case wire shapes. The adapter pattern stays.
- **No content_sha256 invalidation on the backfill** — the SystemPromptBase fill is data-shape normalization, not content change. Existing hashes remain valid by design.
- **Write-side + read-side belt-and-suspenders** — kept BOTH. Read-side is the runtime correctness floor (legacy rows work without any backfill). Write-side prevents new rows from having the issue. Backfill heals at-rest data over time. The doc explains the rationale.
- **`internal/lookup/README.md`** — documents the architectural pattern for future substrates. Adds a "future substrate" contract operators + AI-assisted developers can follow.

### DIFF
--- a/cmd/loomcycle/main.go
+++ b/cmd/loomcycle/main.go
@@ -786,6 +786,17 @@ func main() {
 	} else if n > 0 {
 		log.Printf("mcp_server_defs: backfilled %d rows with content_sha256", n)
 	}
+	// v0.9.x system_prompt_base backfill — fills the field that PR #186's
+	// read-side normalizer derives at runtime. Closes the on-disk data
+	// gap for legacy rows; new rows persist the field directly via the
+	// write-side mergedDef.normalize() (agentdef.go). Idempotent: a
+	// second boot after a full backfill finds zero rows missing the
+	// field. Shares the same bfCtx — bounded boot wait.
+	if n, err := storeIface.BackfillAgentDefSystemPromptBase(bfCtx); err != nil {
+		log.Printf("agent_defs: backfill system_prompt_base partial — %v (rows updated before error: %d)", err, n)
+	} else if n > 0 {
+		log.Printf("agent_defs: backfilled %d rows with system_prompt_base", n)
+	}
 	bfCancel()
 
 	// Memory tool depends on the Store; wire the live backend in now
--- a/internal/api/http/server.go
+++ b/internal/api/http/server.go
@@ -634,7 +634,7 @@ func (s *Server) resolveSkillBodiesForRun(ctx context.Context, agentDef config.A
 			log.Printf("resolveSkillBodiesForRun: SkillDefGetActive(%q) failed: %v", skillName, err)
 			continue
 		}
-		var def skillDefOverlay
+		var def lookup.SubstrateSkillDef
 		if err := json.Unmarshal(row.Definition, &def); err != nil {
 			log.Printf("resolveSkillBodiesForRun: parse %s definition: %v", row.DefID, err)
 			continue
@@ -711,16 +711,6 @@ func (s *Server) emitSystemPromptEvent(
 	if err := s.store.AppendEvent(ctx, runID, "system_prompt", b); err != nil {
 		log.Printf("store: AppendEvent(system_prompt) failed for run %s: %v", runID, err)
 	}
-}
-
-// skillDefOverlay mirrors internal/tools/builtin.skillDefOverlay.
-// Re-declared here so the api/http package doesn't import
-// tools/builtin (which would invert the existing dependency
-// direction).
-type skillDefOverlay struct {
-	Body         string   `json:"body,omitempty"`
-	Description  string   `json:"description,omitempty"`
-	AllowedTools []string `json:"allowed_tools,omitempty"`
 }
 
 // historyPolicyForAgent returns the v0.8.7 Context.history scope policy
--- a/internal/api/http/server.go
+++ b/internal/api/http/server.go
@@ -615,61 +615,44 @@ func (s *Server) resolveSkillBodiesForRun(ctx context.Context, agentDef config.A
 	if len(agentDef.Skills) == 0 || s.store == nil {
 		return agentDef, prov
 	}
-	// Fast pre-check: any DB-active row for any skill name?
-	anyActive := false
-	activeBodies := make(map[string]string, len(agentDef.Skills))
+	// Resolve each skill via the canonical lookup chain (substrate →
+	// static). lookup.Skill returns Source="substrate" when a DB-active
+	// row resolved + Source="static" when falling back to the boot
+	// SkillsRoot bundle.
+	resolutions := make(map[string]lookup.SkillResolution, len(agentDef.Skills))
 	activeDefIDs := make(map[string]string, len(agentDef.Skills))
+	anySubstrate := false
 	for _, skillName := range agentDef.Skills {
-		row, err := s.store.SkillDefGetActive(ctx, skillName)
-		if err != nil {
-			var nf *store.ErrNotFound
-			if errors.As(err, &nf) {
-				continue
-			}
-			// Degrade THIS skill to static; keep checking the rest.
-			// A transient store hiccup on one skill must not drop DB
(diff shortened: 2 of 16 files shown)


denn-gubsky:
### Code-review follow-ups (3 HIGH findings addressed)

Commits 3c23445, 2c6948c, ba7dacc on `feature-static-dynamic-equalization` close the HIGH-rated findings from the parallel code review:

| Finding | Resolution | Commit |
|---|---|---|
| **#1** Backfill skip path missing log line | `log.Printf` added in both sqlite + postgres `continue` branches of `BackfillAgentDefSystemPromptBase`. The doc-promised "skipped with a log line" now actually delivers — operators see broken-JSON rows in the boot log. | 3c23445 |
| **#2** Drift test pinned only one side of the mirror | New `TestMergedDef_DriftDetection_VsLookupSubstrateAgentDef` in `internal/tools/builtin/agentdef_test.go` reflects `mergedDef` directly (it's in-scope there) + compares against `lookup.SubstrateAgentDef` with an explicit exemption list. A future field added to `mergedDef + config.AgentDef` but missed on `SubstrateAgentDef` now fails CI. Existing test's docstring updated to be honest about which direction it covers. | 2c6948c |
| **#3a** `lookup.Skill` was dead production code | `resolveSkillBodiesForRun` refactored to call `lookup.Skill` — ~30 LOC of inline substrate→static fallback deleted. Behaviour preserved (substrate-hit wins, empty-body falls to static, per-skill iteration, provenance map captures only substrate sources). The non-NotFound-error log line moved into `lookup.Skill` so the diagnostic survives. | ba7dacc |
| **#3b** `lookup.MCPServer` couldn't replace `main.go`'s inline lookup (stdio fields) | Docstring + README updated to document the intentional limitation: `MCPServerSpec` is the http / streamable-http subset; stdio's Command/Args/Env stays inline-in-main.go by design. The resolver remains the right call for HTTP-only sites (the future `/ui/library/mcp-servers` GET endpoint). | ba7dacc |

`go test ./... -timeout 300s` green. No verified-clean findings regressed.

The four verified-clean items from the review are unchanged: the static-cfg short-circuit, the `mergedDef.normalize()` vs `lookup.NormalizeAgentDef` duplication (intentional — different types), the `backfillSystemPromptBase` per-package duplication (consequence of separate sqlite/postgres packages), and the concurrency / security / wire-protocol posture.

PR #247, opened 2026-05-27
feat(loop): same-provider retry on retryable errors + mock-stable fallback target

denn-gubsky:
## Summary

Real-world experience: real-provider 429s typically clear in 1-3 seconds, much shorter than the 30s \`MarkRateLimited\` cooldown. The v0.12.x loop fires MarkRateLimited on the FIRST retryable error and immediately escalates — for sustained burst rates this collapses success rate dramatically.

This PR adds a same-provider retry budget that fires BEFORE MarkRateLimited / fallback, so transient 429s self-heal without poisoning the matrix.

## Experimental result (x1000 @ 10% 429)

| | Single-provider config | **Retry + fallback** |
|---|---|---|
| Success rate | 18.5% | **100%** |
| Wall time | 22 min | **~4 min** |
| p99 latency | 3199ms | **3107ms** |
| Provider distribution | mock=1747, unknown=1253 | mock=451, **mock-stable=2549** |

Snapshots side-by-side at \`~/work/loomcycle-internal/load-tests/x1000-429-rate-0.10/\` (broken) vs \`x1000-429-rate-0.10-retry-fallback/\` (this PR).

## What's in this PR

**Engineering — substrate retry:**

- New \`RunOptions.MaxSameProviderRetries int\` field. On retryable error (errclass.Retryable), retry the same (provider, model) up to N times with exponential backoff (100ms / 300ms / 900ms / 2.7s / 8.1s; capped internally at 5) before MarkRateLimited cools the matrix and ReResolve escalates. Both the \`Call()\` error path and the in-stream EventError path get the same treatment. Resets on successful Call() and on tryProviderFallback switches.

- Permanent errors (400 / 401 / 403 / 422 / cancelled) are NOT retried regardless of the budget.

- \`EventRetry\` frames flow to the SSE stream so consumers see the retry live.

**Config plumbing:**

- \`UserTier.RetryAttempts\` yaml field — sits alongside \`fallback_on_error\` + \`max_fallback_attempts\` so all the resilience knobs cluster.

- Wired through all 4 RunOptions construction sites (POST /v1/runs, sub-agent spawn, continue-session, restore).

**Mock-stable provider (for cost-free fallback experiments):**

- \`NewStable()\` factory + stableDriver wrapper that registers as id \`mock-stable\`. Same FSM as the existing mock; failure rates pinned at zero regardless of env.

- Registered alongside \`mock\` when \`LOOMCYCLE_MOCK_ENABLED=1\` is set.

**Test harness:**

- New \`loomcycle.mock-fallback.yaml\` uses per-agent \`models.middle: [{mock,…}, {mock-stable,…}]\` multi-candidate lists. **Critical observation:** the agent.provider/model PIN form leaves the resolver with no alternative when fallback fires — the multi-candidate \`models:\` form gives the resolver somewhere to escalate to.

- New \`run-mock-fallback.sh\` companion wrapper.

## Default behaviour preserved

\`retry_attempts: 0\` (the yaml default for tiers that omit the field) preserves v0.12.x behaviour exactly: the first retryable error fires MarkRateLimited and propagates / falls back immediately. Operators opt into the new resilience.

## Test plan

- [x] 8 new loop tests: retry-then-success, exhausted-then-mark-rate-limited, default-zero-no-retry, permanent-error-no-retry, clamps-at-safety-cap, emits-retry-event, backoff-exponential.
- [x] 2 new mock tests: stable-never-injects, stable-driver-id-overrides.
- [x] \`go test -race ./...\` clean. \`gofmt\` + \`go vet\` clean.
- [x] Live experiment x1000 @ 10% 429 = 1000/1000 success.

## Out of scope

- **Operator-tunable MarkRateLimited cooldown** (today hardcoded 30s). Some providers' 429s clear in seconds; 30s is wasteful. Captured in FINDINGS.md as a follow-up.
- **Per-agent retry override.** Today retry_attempts lives on the user_tier; future iteration could expose agent-level override for high-stakes flows that want 0 retries even under a generous tier.
- **Cluster-mode fallback testing.** Single-binary first.

### DIFF
--- a/cmd/loomcycle/main.go
+++ b/cmd/loomcycle/main.go
@@ -1708,6 +1708,14 @@ type providerResolver struct {
 	// + runtime-fallback paths under load. See
 	// internal/providers/mock/driver.go.
 	mock providers.Provider
+
+	// v0.12.9 — companion `mock-stable` provider. Same driver
+	// shape, failure rates pinned at zero regardless of env. Lets
+	// operators configure `[mock, mock-stable]` as a tier candidate
+	// list with `fallback_on_error: true`, so a 429 on the primary
+	// escalates to the stable variant under tryProviderFallback —
+	// exercising the recovery path against a known-good target.
+	mockStable providers.Provider
 }
 
 // buildEmbedder turns cfg.Memory.Embedder into a constructed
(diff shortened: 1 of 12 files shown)


denn-gubsky:
Rebased onto main (PR #246 merge head 1497795) — local merge against main is clean, but GitHub's mergeable cache is stale at `acdcd2f` instead of the actual remote head `db8a2ca`. Posting this comment to trigger a re-sync.

PR #303, opened 2026-05-31
test(runtime): Schedules + Webhooks + Memory (incl. dedup) runtime suites + 30-min soak

denn-gubsky:
## Why

Runtime (live-binary, over-the-wire) test suites for **Schedules (RFC E)**, **Input Webhooks (RFC H)**, and **Memory (RFC I — built-in memory, all features including dedup)**, plus a **30-minute stability soak**. Complements the `internal/` unit tests (which validate the code) by validating the *wire* against a real loomcycle process, following the existing `test/runtime/<feature>/` convention.

Everything is **deterministic and dependency-free** (mock provider — no real LLM, no API key, no cost), so the suites are safe to run repeatedly and to soak. The one exception (vector dedup) needs Postgres+pgvector and is cleanly gated.

## What

**Stub embedder** (`internal/providers/stubembedder`, `LOOMCYCLE_EMBEDDER_STUB=1`-gated, mirrors the mock provider) — a deterministic signed-hashed-bag-of-words embedder so near-duplicate text → cosine ≥ the 0.92 dedup threshold. Makes dedup/search assertable without a real embedder; invisible to a production binary that doesn't set the flag. + unit tests.

**`mock-mem-search`** driver model — issues one `Memory.search` with search-time dedup, so the vector suite can drive dedup through a real agent run.

**Runtime suites** (`test/runtime/`):
| Suite | Coverage |
|---|---|
| `schedules/` | ScheduleDef create→get→`/names`→**real cron fire** (`* * * * *`, tick=1s, ~60s)→completed run→retire→quiesced |
| `webhooks/` | WebhookDef CRUD; signed `sha256=` delivery→202+run; bad-sig→401; oversized→400; **valid-sig replay→deduped, no dup run**; rate-limit→429 |
| `memory-core/` | admin K/V CRUD (user+agent scopes); MemoryBackendDef CRUD incl. the **empty-`prefix_pattern` tenancy rejection** + **non-http(s) `base_url` SSRF rejection** (the fixes from #302) end-to-end |
| `memory-vector/` | PG+pgvector+stub: embed-on-write, `embed_stats`, **search-time dedup collapse** (`dedup_dropped≥1`, distinct row survives), reembed dry_run. Skips without `LOOMCYCLE_TEST_PG_DSN` |
| `stability/` | **30-min soak**: continuous runs + signed webhooks + bounded memory churn + cron fires; samples RSS/ops/errors/fires; asserts no crash, err≤1%, RSS≤1.8× warmup (leak guard), sweeper liveness |

`make runtime-mock` (3 fast suites) / `make runtime-vector` / `make runtime-soak`. None added to the default `test`/CI gate (they boot a live binary; one waits ~60s for a cron fire).

## What was tested (executed locally)

- `schedules/` **PASS** — real cron fire observed at ~57s → completed run.
- `webhooks/` **PASS** — all six paths.
- `memory-core/` **PASS** — incl. the #302 tenancy + SSRF rejections over the wire.
- `memory-vector/` **PASS** against a live **pg17 + pgvector 0.8.2** (installed locally; no docker) — `dedup_dropped=1`, the near-duplicate cluster collapsed while the distinct row survived. Skip path also verified.
- `stability/` smoke **PASS** at 120s (0 errors, flat RSS ~32–39 MB, scheduler fired→completed, ~9k ops). **The full 30-min run is executing now; I'll post the result.**
- `go build/vet/test ./...` clean; `gofmt` clean.

## Discovered (documented, not changed here — a wire-contract call for the maintainer)

The webhook **replay guard returns 401 `unauthorized`** (`server.go:195`, deliberately "opaque, same as a sig failure"). That is arguably misleading for a valid-but-duplicate delivery (a 200/409 idempotent ack is the usual contract; the replay path is only reachable *after* signature verification, so no unauthorized attacker is being shielded). The `webhooks/` suite asserts the shipped behaviour so any future change is conscious.


denn-gubsky:
### 30-minute stability soak — PASS ✓

Ran `LOOMCYCLE_SOAK_SECONDS=1800 ./test/runtime/stability/run.sh` to completion:

```
1800s soak: alive, err=0.000%, rss 29872→51984KB (<1.8x),
            sched fired→completed, ops=131271
```

- **131,271 ops, 0 errors** (HTTP runs + signed webhooks + bounded memory churn), ~86k completed runs.
- **RSS bounded** — GC sawtooth, warmup 29.9 MB → final 52 MB (1.74×, under the 1.8× leak guard). No upward trend.
- **Scheduler liveness** — the `* * * * *` cron kept firing→completing across the full 30 min (sweeper never died).
- **No crash.**

memory-vector also executed against a live local pg17 + pgvector 0.8.2 (`dedup_dropped=1`). All other suites green. CI green.

PR #307, opened 2026-06-01
feat(rfc-j): code-js synthetic provider — stateless replay model (supersedes #306)

denn-gubsky:
## What this is

The **replay / durable-execution** implementation of the RFC J code-js provider — the reviewed successor to PR #306 (Mechanism-1 parked-goroutine model). Same feature, cleaner execution core. **Supersedes #306** (close that one); this is off `main` and self-contained (not stacked). Decision recorded in RFC J Appendix B.

## Why replace the Mechanism-1 core

PR #306's review surfaced that Mechanism 1 held a live `goja.Runtime` + parked goroutine **per in-flight run across every dispatch gap**, made code-js the only *stateful* provider (violating the documented "stateless across calls" contract), and was **not resumable across restart/replica** (a v0.9.x HA blocker). I measured the costs (goja, real numbers): ~10 KB/parked run floor (scales with live JS state), ~64 µs/turn to reconstruct (negligible vs a tool dispatch). The replay model removes the whole class of problem.

## The replay model

The provider is **stateless**. Each `Provider.Call` builds a fresh runtime, **fast-forwards the tool results already recorded in `req.Messages`** (the transcript *is* the durable memoization log), and stops at the first un-recorded call (the "frontier") via `rt.Interrupt` — which the loop dispatches, then re-invokes `Call` with the result appended. The runtime is discarded within each `Call`; nothing is held across the dispatch gap. Loop-driven dispatch (Decision 5) is unchanged — the provider still only emits `EventToolCall` and never imports `internal/tools`.

What this buys (vs #306): honors the stateless-Provider contract; **deletes** the continuation registry, the parked goroutine, the leak-backstop watcher, the token-keyed lookup; cancel is just the Call's `ctx` (no goja `#97` parked-host-func subtlety); **resumable across restart/replica for free**; each tool dispatches exactly once.

## Ambient determinism — always on, so replay can't diverge

Replay requires re-execution to reproduce every value the JS reads. Tool results do (transcript). The remaining sources — `Date.now()` / `Math.random()` / `new Date()` — are **hooked in `sandbox.go`**: `Math.random()` seeded from a per-run seed; `Date.now()` = a real run-start **anchor** + a monotonic per-call offset. A no-I/O sandbox with the clock+RNG hooked is fully deterministic given `(transcript, seed, anchor)`, so **divergence is impossible by construction**, not merely detected. Seed/anchor come from `RunMeta` (`RunID`, `StartedAt`, stamped by the loop). `LOOMCYCLE_CODE_AGENTS_DETERMINISTIC=1` sharpens to "freeze anchor+seed across runs" (cross-run reproducibility). A name mismatch against the recorded sequence still fails loud (`code_agent_replay_divergence`) as a defense-in-depth guard.

This is the refinement from the design discussion: instead of recording every `Date.now()`/`random()` value (which would need a durable side-log and reintroduce per-run state), we record the *minimum to regenerate* — a seed + a single anchor, both derived from values the run record already persists. Storage-free and restart-safe.

## What was tested

- **`internal/providers/codejs` (race-clean):** multi-turn replay with the recorded get-result flowing into the later set value (exactly-once dispatch); `IsError`→catchable throw; uncaught throw→`EventError`; allowed-tools `ReferenceError`; `eval`/`Function` blocked; CPU-bound loop killed by `Interrupt`; cross-run reproducibility (`--deterministic`); **always-on ambient determinism stable across replays** (anchored `Date.now`, seeded `Math.random`); **replay-divergence fails loud**; 24 concurrent isolated runs; compile hash.
- **Real-loop integration:** `loop.Run` drives the provider through tool calls dispatched via the loop's own `Dispatcher.Execute`.
- **Manual HTTP e2e** (`POST /v1/runs`): a `memory.get→set→return` code-agent replays correctly through the real loop; memory persists across runs (`c=0` then `c=1`); `Date.now()` returns the real run-start anchor (`t=1780286744384`); `Math.random()` is per-run seeded; `usage.model=loomcycle/code-js`, zero tokens, `end_turn`.
- Full gate clean: `gofmt -l`, `go build/vet/test ./...` (`-race` on codejs/loop/config).

## Fixed vs #306

A latent naming bug (also in #306, masked because #306's smoke agent never *called* a tool): `allowed_tools` matching is case-sensitive **canonical** (`policy.Matches`), so the docs/example's lowercase `[memory, channel]` silently omitted the tools → `ReferenceError`. Corrected to canonical `[Memory, Channel, …]` + `memory_scopes`, with the naming contract (canonical `allowed_tools` vs lowercase JS objects) documented in the help topic.

## Deferred (unchanged from #306's scope)

10k stress test (now also: the O(N²) pure-JS re-execution bound for many-tool-call agents — measured ~N²/2 segment-runs); scheduler/webhook/A2A e2e harness tests; `sql-query`/`format-converter`/`router` examples; OTEL `provider.code_hash` span attribute (hash computed + exposed via `Compile`, recorder.go wiring not done).


internal/help/builtin/code-agents.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Sharp edges
+
+- **Glue-logic fast, not data-processing fast.** goja is interpreted (no
+  JIT). CPU-bound work over megabytes belongs in an MCP server. The
+  design center is the ~100ms-budget glue step.
+- **One tool call at a time.** Parallel tool calls within one code-agent
+  are out of v1 (the suspend point is one-at-a-time). Fan out with
+  `agent.spawn(...)`, which is already concurrent at the loomcycle layer.
+- **Not resumable across a restart.** A code-agent's run state is live
```
denn-gubsky: This should be refactored since we just added resume capability for scripts

denn-gubsky (commented):
Change the Context internal help artcile stating JS code continuation and resumability

internal/help/builtin/code-agents.md
denn-gubsky: Fixed in 1544c8e. Reversed the sharp edge: with the replay model a code-agent holds no in-process state, so a run interrupted mid-flight resumes from the transcript (the win that motivated replay). Also rewrote the Determinism section (within-run determinism is always-on — seeded Math.random + anchored Date.now — which is *why* replay is resumable) and dropped the remaining parked-continuation wording ("suspend point", "suspends then resumes").

PR #319, opened 2026-06-01
fix+feat: un-red main, exempt code-js from MaxIterations, drop dead WebhookDefScopes

denn-gubsky:
Follow-up after the v0.16.0 QA pass (PRs #310–#318). Three concerns, one commit each.

## 1. `f35bc09` — un-red main (parallel-merge artifacts) ⚠️

**main is currently red.** Merging the 9 QA PRs in parallel produced two artifacts the per-PR CI couldn't catch (each was green against its own base):
- `codejs_test.go`: the squash dropped the closing braces between two adjacent test funcs → the package **didn't compile** (`expected '(', found TestCodeJS_ReplayInputDivergence_FailsLoud`).
- `internal/otel/recorder.go`: a new attr key (#314) shifted a `const` block → **gofmt dirty**.

Both formatting/boundary-only. This commit alone restores `go build/vet/test ./...` + gofmt green — fast-track it if you want main un-red before reviewing the rest.

## 2. `5d95621` — code-js exempt from MaxIterations; the run timeout is the bound

Your ask #1. For a code-agent each loop turn is one internal tool-dispatch step of a single `run()`, so the `MaxIterations` soft-cap (16) is unusable — it silently truncates `run()` after 16 sequential calls.
- New `providers.Capabilities.UnboundedIterations` (code-js → true); the loop honors it (capability-driven, **not** ID-coupled). LLM drivers unchanged — `MaxIterations` still guards runaway tool-use. A `1<<20` hard ceiling remains as a pure runaway backstop.
- **Made the code-js timeout a whole-run deadline** (not a fresh per-turn one): `interruptWatch`'s budget is the run's *remaining* wall-clock, derived from the stable `RunMeta.StartedAt` — so the sum across replay turns can't exceed `RunTimeout`. This is what makes disabling the cap safe (your "timeout is enough" is now literally true at the run level).
- Tests: a loop unit test (cap on → 25-call run completes `end_turn`; off → caps at 3), and the `code-js-stress` runtime suite updated — **[A]** 25 calls complete with 25 incrs (no cap), **[A2]** a *new* runaway agent (infinite tool-call loop) is cut by a 5s run timeout (error event, `stop≠end_turn`, returns in ~5s — not hung). `runtime-mock` (all suites) green.

## 3. `0b252a9` — remove dead `WebhookDefScopes` (the deferred fix)

Your ask #2 (applicable deferred items). #316 removed the identical dead `MemoryBackendDefScopes` and explicitly tracked `WebhookDefScopes` as the same defect for a follow-up — this is it. Parsed-but-never-read agent-yaml field on an admin-only tool; deleted + the two misleading error strings reworded.

## Deferred items I consciously kept deferred (with rationale)

- **A2A gRPC inbound tenant** → needs RFC L's authoritative tenant (v1.1.0); nothing to enforce yet.
- **A2A gRPC SSRF residual (rebinding)** → #317 already blocks literal IPs; the rebinding fix needs a TLS-preserving custom dialer (the QA agent's named "TLS-downgrade risk") — rushing it in a stabilization window is net-negative. Belongs in the A2A-hardening follow-up.
- **Channel L2 idempotency** → v1.1 HA window (design captured).
- **RFC-K Finding G (inert TenantID)** → RFC L.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,16 +65,19 @@ jobs:
           go-version: "1.26.x"
           cache: true
 
-      # The code-js runtime suites are deterministic (synthetic provider, no
-      # LLM / key / Postgres) and assert storage state via the sqlite3 CLI +
-      # parse SSE with python3. Both are preinstalled on ubuntu-latest; this
-      # makes the dependency explicit and fails fast if the image changes.
+      # The FUNCTIONAL code-js runtime suite is deterministic (synthetic
+      # provider, no LLM / key / Postgres) and asserts storage state via the
+      # sqlite3 CLI + parses SSE with python3. Both are preinstalled on
+      # ubuntu-latest; this makes the dependency explicit and fails fast if the
+      # image changes. NOTE: the STRESS suite (`make runtime-stress`) and the
+      # soak/load suites are intentionally NOT run here — load, stress, and
+      # sustainability tests run on the operator's machine, not in CI.
       - name: Ensure sqlite3 + python3
         run: |
           command -v python3 >/dev/null || { echo "python3 missing"; exit 1; }
           command -v sqlite3  >/dev/null || sudo apt-get update && sudo apt-get install -y sqlite3
 
-      - name: Runtime code-js suites
+      - name: Runtime code-js suite (functional only)
         run: make runtime-codejs
 
   ts-adapter:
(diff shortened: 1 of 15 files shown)


denn-gubsky:
Added commit `3bec798` per a follow-up request: **stress/soak/load suites now run locally only, not in CI.** The CI `runtime-codejs` job runs the functional `code-js` suite only; the stress suite (concurrency + the new runaway-timeout test) moves to a local-only `make runtime-stress`. No coverage lost — the iteration-cap exemption is also unit-tested (`TestRun_UnboundedIterations_ExemptFromMaxIterations`), which stays in CI.

PR #322, opened 2026-06-01
feat(api): add POST /v1/_resolve/probe for operator-triggered re-probe

### LINKED ISSUE #88: Add admin endpoint POST /v1/_resolve/probe to trigger immediate resolver re-probe
## Problem

The resolver probes each provider's \`/v1/models\` endpoint at a fixed interval (\`LOOMCYCLE_RESOLVE_PROBE_INTERVAL_MS\`, default 15 min). When a transient network blip hits during a probe (DNS hiccup, brief upstream outage, VM losing egress for 30s), **every** external provider gets marked stalled, and production 503s for up to a full probe interval until the next tick.

Surfaced 2026-05-13: ~5-min network outage on the production VM at ~09:37 UTC caused \`cv-batch-adapter tier=low\` to 503 with \`no provider available for requested tier (no reachable provider with a non-stalled model)\` for ~15 minutes, even after network connectivity was fully restored at ~09:38. Operator had no way to unstick this short of \`systemctl restart loomcycle\` (drops in-flight runs).

## Proposed solution

New admin endpoint, bearer-authed like the v0.8.6 \`POST /v1/_channels/_system/...\` endpoint:

\`\`\`
POST /v1/_resolve/probe
Authorization: Bearer \$LOOMCYCLE_AUTH_TOKEN
\`\`\`

Triggers an immediate probe of every configured provider. Returns the same probe-result shape that's currently emitted to the log (per-provider reachable/excluded/unreachable status). Useful for:

- Operator unsticking after a transient network outage (the surfacing case)
- Health checks that want to actively verify provider connectivity (alternative to \`GET /healthz\` which only checks self)
- Post-deploy / post-config-change validation before serving traffic

## Why this matters

Probe interval is a tradeoff between staleness (long interval → slow recovery from transient outages) and probe-call frequency (short interval → more \`/v1/models\` calls). Default 15 min is reasonable steady-state. The endpoint provides operator-controlled override for the exceptional cases without forcing a runtime restart.

Operator-side workaround we shipped today: drop default to 5 min via \`LOOMCYCLE_RESOLVE_PROBE_INTERVAL_MS=300000\`. Bounds recovery window to ≤5 min. Still not instant — the admin endpoint would close that gap.

## Implementation scope

- New HTTP route in \`internal/api/http/server.go\` mirroring the system-channels admin endpoint pattern (bearer-authed, JSON request/response).
- New method on the resolver (probably \`Probe(ctx)\` or \`ProbeNow(ctx)\`) that runs the existing probe goroutine's body inline and returns the result map.
- Existing probe-tick code in \`cmd/loomcycle/main.go\` (line ~597 area) becomes one caller; the new endpoint is the second caller.
- Same shape as the channels admin endpoint: 200 with per-provider result, 401 on bad bearer, 500 on internal probe error.

## Followups not in scope here

- Reducing the default probe interval below 15 min — depends on provider \`/v1/models\` rate limits; safer as operator opt-in.
- OpenTelemetry export of probe results — part of the v0.9.x OTEL sweep.

## Surfacing context

- v0.8.10 deployed to production VM 2026-05-13
- Production hit the issue ~30 min later when a transient VM-network outage stalled every provider
- Existing workarounds: wait for next probe (15 min worst case), restart loomcycle (~30s but drops in-flight runs), tune \`LOOMCYCLE_RESOLVE_PROBE_INTERVAL_MS\` (down from 15 → 5 min, applied today)

denn-gubsky:
Closes #88.

## Why

The resolver re-probes each provider's `/v1/models` on a fixed interval (`LOOMCYCLE_RESOLVE_PROBE_INTERVAL_MS`, default 15 min). When a transient network blip hits *during* a probe — DNS hiccup, brief upstream outage, the VM losing egress for ~30s — **every** provider gets marked stalled, and production 503s with `no provider available for requested tier` for up to a full interval until the next tick, even after connectivity is restored. The only recovery paths were waiting it out or `systemctl restart loomcycle` (which drops in-flight runs). This surfaced in production on 2026-05-13.

## What

A bearer-authed admin endpoint — `POST /v1/_resolve/probe` — that triggers an **immediate, synchronous** re-probe and returns the refreshed availability matrix in the same wire shape as `GET /v1/_resolver`.

The probe mechanism itself already existed: `Resolver.ForceProbe` landed with the snapshot-restore work, but its only caller was the restore handler. This exposes it as a standalone operator trigger.

Design choices:
- **Per-provider failures are data, not errors.** A provider still unreachable after the probe comes back as `reachable: false` + `last_error` inside a **200** — same as the periodic sweep records it. This mirrors the GET handler's philosophy.
- **503 only when it can't probe at all:** `resolver_unavailable` (degraded startup, no resolver) or `probe_unavailable` (no probe loop wired, e.g. `--no-http`).
- Added `Resolver.ForceProbeWired()` so the endpoint returns that honest 503 rather than a 200 over a matrix it never actually refreshed (`ForceProbe` is a silent no-op when unwired).

Known characteristic (not addressed — operator-only endpoint): two simultaneous POSTs run the probe twice (redundant upstream calls), but matrix writes are mutex-serialized so there's no corruption — no worse than the existing periodic-loop ↔ restore race.

## What was tested

**Unit** (`internal/api/http/resolver_test.go`):
- `TestHandleResolveProbe_TriggersForceProbeAndReturnsRefreshedMatrix` — the core #88 behaviour: the force-probe callback runs exactly once and the response reflects the post-probe state (deepseek flipped excluded → reachable).
- `TestHandleResolveProbe_503WhenNoResolver`, `TestHandleResolveProbe_503WhenProbeNotWired`.

**Manual** (built + booted with a minimal mock config):
- `POST` with good bearer → **200** with all `last_check` timestamps stamped at request time (proves a real fresh probe ran, not a cached read).
- no bearer / bad bearer → **401**; `GET` on the POST-only route → **405**.

`go test ./...` and `gofmt` clean.


denn-gubsky:
**Follow-up commit `1eaace0` — exposed across all transports for completeness.**

The first commit added the HTTP endpoint as a direct handler. This one threads it through the `connector.Connector` contract so every wire surface reaches it, per the architecture (HTTP implements; gRPC + MCP consume; adapters mirror):

- **Connector**: `ResolveProbe(ctx) (ResolverMatrix, error)` + DTOs + two typed sentinels (`ErrResolverUnavailable`, `ErrResolveProbeUnavailable`). The HTTP handler was **refactored to delegate** to this canonical method (one probe-and-snapshot code path; identical wire bytes since `ResolverMatrix`'s JSON tags match `GET /v1/_resolver`).
- **gRPC**: `ResolveProbe` rpc + `ResolverMatrixResponse` messages (`proto/loomcycle.proto`, regenerated with the pinned `protoc-gen-go v1.36.11` / `protoc-gen-go-grpc v1.6.2`); handler maps both sentinels to `codes.Unavailable`.
- **MCP**: `resolve_probe` meta-tool (`tools/list` → 39).
- **TS client**: `client.resolveProbe()` + `ResolverMatrix` types; package `0.15.0 → 0.16.0`.
- **Docs**: CONFIGURATION cross-surface note; corrected a stale `33 meta-tools` → `39` in MCP_SERVER.md.

New tests on every surface (gRPC dispatch + Unavailable mapping; MCP `tools/call` dispatch; TS POST-shape + 503→`UnavailableError`); `go test ./...`, `go vet ./...`, gofmt, `tsc --noEmit`, and the 178-test TS suite all green.

PR #334, opened 2026-06-02
feat(authz): tenant-scoped read boundary + /v1/_me (UI multi-tenant PR 1/2)

denn-gubsky:
**PR 1 of 2** for the Web UI multi-tenant authorization feature. This is the **backend half** — it makes "a tenant sees only its own workspace" a real **server-side** boundary (the confirmed decision: not UI hiding) and gives the SPA a role source. **No UI change here** (PR 2).

## Why

Post-RFC-L the middleware resolves an authoritative `Principal{tenant, subject, scopes}`, but the read endpoints didn't filter by it — a non-admin token could read another tenant's runs/agents via the API. And the SPA had no way to learn its own role (super-admin vs tenant).

## What

- **Schema:** denormalize `tenant_id` onto `runs` (sessions had it; runs didn't, so tenant-scoped lists needed a JOIN — and the UI polls them). Postgres migration `0036` (ALTER + backfill from sessions + index); idempotent SQLite ALTER+backfill+index; `store.RunIdentity`/`store.Run` gain `TenantID`; `CreateRun`/`scanRun`/`runColumns` thread it (both backends); set at all four run-creation sites from the run's effective tenant.
- **`GET /v1/_me` (whoami):** `{tenant_id, subject, scopes, is_admin, legacy}` from the ctx principal (open mode → synthetic admin so the dev UI works). Exempt from the `/v1/_*` admin gate — any authenticated principal needs its own identity.
- **Enforcement** (`auth_principal.go`): `principalTenantScope` (list reads) + `tenantVisible` (single-row reads), mirroring `applyPrincipal`'s posture — admin/open sees all + can focus a tenant via `?tenant=`; a tenant principal is forced to its own tenant (wire/`?tenant` ignored). Applied to `GET /v1/users/{user_id}/agents` (cross-tenant runs filtered out) and `GET /v1/agents/{agent_id}` (cross-tenant → opaque 404). Transcripts are already gated (per-subject) by the QA-fix `sessionOwnershipOK`.

## Tested

whoami (admin/tenant/open); `principalTenantScope` + `tenantVisible`; **cross-tenant list isolation** (a tenant token listing a user with runs in two tenants sees only its own; admin sees both); store contract + existing http suites still green; `go build/vet/test ./...` + gofmt clean. Manual: legacy admin → `/v1/_me` `is_admin:true`; minted `acme/alice` token → `is_admin:false, tenant_id:acme`.

## ⚠️ One decision to settle for PR 2 (flagged)

There's a **within-tenant granularity seam**: the merged QA-fix gates **sessions/transcripts per-SUBJECT** (`sessionOwnershipOK` = tenant AND subject), while this PR's **run/agent reads are per-TENANT** (the approved UI role model: admin=all-tenants vs tenant=own-tenant). Both are cross-tenant-safe; they differ on whether a tenant token sees its *whole tenant* or *just its own subject*. PR 2 will reconcile the view granularity — worth your steer: **should a "tenant" role see the whole tenant's workspace, or only its own subject's?** (My PR-1 run/agent enforcement assumed whole-tenant per the plan; the QA fix assumed subject. I left the QA fix untouched since loosening it is a security change.)

## Next (PR 2)
Frontend: `/login` token-entry page, `usePrincipal()` via `/v1/_me`, 401→login, role-aware nav (tenants don't see admin tabs), super-admin tenant switcher; `make build-ui`.

### DIFF
--- a/internal/api/http/auth_principal.go
+++ b/internal/api/http/auth_principal.go
@@ -280,25 +280,28 @@ func (s *Server) applyPrincipal(ctx context.Context, wireTenant, wireUser string
 }
 
 // sessionOwnershipOK reports whether the ctx principal may continue or read
-// sess (RFC L cross-principal boundary). A continuation or transcript read
-// runs under / exposes the SESSION'S stored (tenant, subject) — so it must
-// belong to the same authoritative principal that created it. Session ids are
-// NOT secrets (they are returned to the caller, logged, shown in the UI, and
-// embedded in events/transcripts), so without this check principal-A could
-// POST to principal-B's session id and have the run execute under B's tenant
-// + subject: cross-tenant memory read/write, replay of B's transcript back to
-// A, fairness-cap evasion, and attribution as B.
+// sess. A continuation runs under / a transcript read exposes the SESSION'S
+// history, so the gate keeps a caller from acting on a session OUTSIDE ITS
+// TENANT — session ids are NOT secrets (returned to callers, logged, shown in
+// the UI, embedded in transcripts), so without this a token from tenant-B
+// could POST to a tenant-A session id and get cross-tenant memory read/write,
+// transcript replay, fairness-cap evasion, and attribution in A.
 //
-// Exempt (return true): no principal (open dev mode), and the single-operator
-// legacy principal (Legacy=true) — under the legacy shared secret there is
-// exactly one principal, so there is no cross-principal boundary to cross, and
-// pre-RFC-L sessions carry default/empty identity that would otherwise 404.
+// WHOLE-TENANT model (the chosen Web-UI authz granularity): the boundary is
+// the TENANT, not the subject — subjects within one tenant share the tenant's
+// workspace (they collaborate), so any acme subject may read/continue any acme
+// session. The cross-TENANT boundary (the actual security property) stays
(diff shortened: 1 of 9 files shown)


denn-gubsky:
**Follow-up commit `dcd863d` — settled the granularity seam: whole-tenant** (per your call).

Loosened `sessionOwnershipOK` from the QA-fix's per-subject check (`tenant AND subject`) to **tenant-only + super-admin bypass**:
- cross-**tenant** continuation/read → still opaque 404 (the QA fix's actual security property is fully preserved);
- same-tenant **different** subject → now allowed (whole-tenant collaboration), so it matches this PR's tenant-level run/agent filtering — the boundary is now consistent across *all* read + continuation paths;
- `substrate:admin` → sees/acts across all tenants (also fixes a latent bug where an admin token was pinned to its own tenant+subject).

This is a deliberate, reviewed loosening of a merged QA fix; the cross-tenant boundary is unchanged. Tests updated (ownership matrix + the continuation test, renamed `RejectsCrossTenantSession`, now asserts a same-tenant different-subject caller passes). `go test ./internal/api/http/...` + gofmt clean.

PR 1 is now internally consistent at the whole-tenant granularity. PR 2 (frontend) will build the role-aware UI on this.

PR #349, opened 2026-06-03
feat(codejs): inline code_body ingestion via AgentDef (no host FS bind)

denn-gubsky:
## Why

A code-js agent's JS could only be read from `agent_code/<name>/index.js` on the sidecar's disk. Every *other* agent attribute — `system_prompt`, `allowed_tools`, `tier` — ingests through the dynamic **AgentDef** substrate (versioned `agent_defs` rows, lineage, `content_sha256`, snapshot round-trip), but the JS body could not. So a code agent couldn't be registered across container boundaries or in a pure-cloud deploy **without binding the container to a host filesystem**.

This closes that asymmetry: a code agent's body now ingests through AgentDef (and yaml `code:`) exactly like its prompt — full symmetry, no host FS bind.

## What

**Substrate + hash**
- `agents.AgentContent` gains `code_body` (`omitempty`, alphabetical tag order). An empty body serialises away ⇒ **every existing non-code row, and every filesystem-backed static code agent, hashes byte-for-byte as before** (zero `content_sha256` churn). Inline code *is* content, so two bodies hash differently (verify/dedup stays correct).
- Field threads the four mirrors: `config.AgentDef.Code` (`yaml:"code"`), `mergedDef`, `lookup.SubstrateAgentDef` + `ToConfigDef`, `AgentContent`.

**Threading (provider stays a leaf)**
- Rides `providers.RunMeta.CodeBody` from the loop to the provider, so `internal/providers` keeps its no-store/no-tools invariant. All four run-creation paths populate it — `RunOnce`, `POST /v1/runs`, messages-continuation, and `runSubAgent` — so sub-agents inherit the body identically.

**Provider + cache**
- The provider prefers an inline body over the filesystem fallback.
- The compiler cache is re-keyed by **content sha256** (not agent name), so a new AgentDef version shipping new JS compiles fresh instead of serving a stale program (and the old by-name "edit-without-restart" sharp edge no longer applies to distinct bytes).

**Gate (reuse the existing switch)**
- No new flag: `LOOMCYCLE_CODE_AGENTS_ENABLED` (the switch that registers the provider) gates inline. `create`/`fork` refuse a non-empty `code_body` when code-js is disabled, enforce a dedicated `LOOMCYCLE_AGENT_DEF_MAX_CODE_BYTES` cap (256 KB default), and parse-compile via a new pure `codejs.Validate` (single source of truth for the compile flags; verified cycle-free).

**TS client (`@loomcycle/client` → 0.19.0)**
- Typed `AgentDefOverlay.code_body` + `ensureCodeAgent({name, code, allowedTools?, …})` convenience, mirror of v0.18.0 `ensureMcpServer`. `AgentDefRowResponse` gains `deduplicated`. Publish is the standard tag-driven step after merge.

## What was tested

Fail-before regression tests (provider precedence verified fail-before by neutering it):
- **codejs**: `TestProvider_InlineBodyExecutesOverFilesystem`, `TestProvider_EmptyBodyFallsBackToFilesystem`, `TestProvider_InlineBodyNeedsNoFilesystem`, `TestCompiler_CacheKeyedByContentHashNotName`, `TestValidate_AcceptsAndHashes` / `_RejectsSyntaxError`.
- **agents/sign**: `TestSign_EmptyCodeBodyHashesIdenticallyToLegacy` (backward-compat), `TestSign_CodeBodyChangesHash`.
- **agentdef**: `TestAgentDefTool_InlineCodeRefusedWhenCodeJSDisabled`, `_RejectsSyntaxError`, `_RejectsOversize`, `_CreateAndForkInherits`; drift test extended.
- **TS**: `ensureCodeAgent` create-payload + dedup `changed=false` (189 tests pass).
- `go test ./...`, `go vet`, `gofmt`, `make build-all`, `npm run typecheck && npm test` all clean.

## Out of scope (noted in code)

- Forking a **filesystem** code agent: fails loud with a hint to supply an inline `code_body` (auto-materialising the disk body would re-introduce the FS dependency we're removing). Forking an **inline** parent inherits the body for free.
- The npm publish of `@loomcycle/client@0.19.0` (tag-driven release step after merge).


denn-gubsky:
## Self-review pass (high-effort, 7 finder angles)

Ran a code-review session against this branch. Surfaced 10 findings; **fixed the 3 highest-impact in `6646e73`**, deferring 2 substrate-drift items to a follow-up PR.

### Fixed in this PR
- **#1 boot-fatal (HIGH):** `validateCodeAgents` validated only the filesystem, so a static yaml `provider: code-js` agent with an inline `code:` body and no host `index.js` — the headline no-FS-bind case — crashed the boot. Now validates the inline body via `codejs.Validate`, mirroring the run-time inline-over-FS precedence.
- **#2 per-turn disk read (HIGH):** the compiler refactor dropped `load()`'s early by-name cache check, so the FS path re-read + re-hashed `index.js` on every replay-turn `Provider.Call`. Restored a by-name `fsCache`; inline stays content-hash-keyed (version-correct) and re-hashes per turn (cheap, no I/O).
- **#3 false dedup contract (MED):** `AgentDef.execCreate` had no content-addressed dedup (unlike `MCPServerDef`), so the TS `ensureCodeAgent` `changed` flag was always `true` and the "byte-identical re-register is a no-op" claim was false. Added the same idempotent-create dedup (`deduplicated=true` against the active row's hash).

Both behavioral fixes have fail-before regression tests (verified by reverting each).

### Deferred to a follow-up PR (substrate-drift class — review #4/#5)
- **`FromYAMLAgent` omits `AgentContent.CodeBody`** (and `agents.Agent` has no `Code` field): the new hash-significant field has only 1 of its producers updated → a code agent hashed via the yaml/.md/CLI path would mismatch the substrate hash. Needs a 3-way hash-equivalence test (per the project's substrate-drift rule).
- **`mergeAgentDef` drops `override.Code`:** an inline `code:` set in the yaml override layer is silently dropped for `.md`-discovered agents.

### Acknowledged, lower-priority (not blocking; worth a look)
- Unbounded inline cache growth under fork/versioning (content-hash key, no eviction).
- No run-time enabled-gate in `Provider.Call` (relies on boot registration == flag; brittle across snapshot-restore / HA replicas).
- `MaxCodeBytes` (256 KB) shadowed by `MaxDefinitionBytes` (128 KB) in the 128–256 KB band.
- `codejs.Validate` duplicates the compiler's compile+hash (drift risk vs the "single source of truth" comment).

PR #357, opened 2026-06-03
feat(triggers): WebHook/Scheduler metadata sourcing + WebhookDef creds parity (PR 2/2)

denn-gubsky:
**PR 2 of 2** (PR 1 #356 — the run-input metadata channel + dual-agent delivery — is on `main`). This sources both data domains at the WebHook/Scheduler def level and wires them into the spawned run. Off `main`, **not stacked**.

## Non-secret metadata (the gap PR 1 created the channel for)

- `config.Webhook` + `config.ScheduledRun` gain a `metadata` field (operator/def-authored → **TRUSTED**), threaded through `merged*Def` / `Substrate*Def` / `ToConfigDef` + the scheduler's `scheduleDef`.
- The **webhook receiver** sets `RunInput.Metadata` from the static def, and routes the previously-**discarded** `payload_mapping` `run_metadata.*` targets into `RunInput.PayloadMetadata` (**UNTRUSTED** — projected from the signed inbound body; fenced downstream).
- The **scheduler** threads `def.Metadata` (no inbound body → no `PayloadMetadata`). Per-fork metadata (a repo per fork) falls out of the overlay naturally.

## Secure-domain parity (user-approved)

- `WebhookDef` gains the fork-time explicit `user_credentials` map (parity with `ScheduleDef`). Receiver credential precedence is now **env-resolved < fork-time `user_credentials` < payload-projected `user_credentials.*`** (the live per-delivery token still wins). Documented that a literal secret in the def is the weaker posture vs env/payload.

> **Secure-domain verification (your original ask):** the RFC F per-user path is confirmed correct + multi-user as-is — env-resolved + per-delivery-payload-projected bearer tokens, isolated per fired run, substituted at the MCP transport, inherited by sub-agents. This PR only *adds* the fork-time map; it changes nothing about the multi-user isolation.

## Persistence

Both fields live in the def `Definition` JSONB → round-trip through snapshot. WebhookDef/ScheduleDef have **no** `content_sha256` (no verify op, RFC E/H), so there is no hash to churn. Updated the `SubstrateWebhookDef` + `SubstrateScheduleDef` drift `want` maps (the conscious-addition gate).

## What was tested

Fail-before (these fields/paths did not exist):
- webhook `TestBuildRunInput_MetadataTrustedAndPayloadUntrusted` (static def → `Metadata`; `run_metadata.*` → `PayloadMetadata`) + `_UserCredentialsPrecedence` (env < fork < payload).
- scheduler `TestBuildRunInput_ThreadsDefMetadata`.
- `TestWebhook_DriftDetection` + `TestSchedule_DriftDetection` updated.
- `go test ./...`, `go vet`, `gofmt` clean.

## End-to-end (after merge)

Define a webhook with `metadata: {review_policy: strict}`, `payload_mapping: {"run_metadata.repo": "$.repository.full_name", "user_credentials.slack": "$.auth.slack"}`. A signed GitHub-shaped POST → a code-js agent reads `input.metadata.review_policy` (trusted) + `input.payload_metadata.repo` (untrusted) + the Slack token substituted via `${run.credentials.slack}`; an LLM agent's transcript shows the trusted metadata block + the fenced `<run_metadata>` block.

### DIFF
--- a/internal/api/webhook/metadata_test.go
+++ b/internal/api/webhook/metadata_test.go
@@ -1,0 +1,59 @@
+package webhook
+
+import (
+	"testing"
+
+	"github.com/denn-gubsky/loomcycle/internal/config"
+)
+
+// TestBuildRunInput_MetadataTrustedAndPayloadUntrusted pins the non-secret
+// metadata sourcing: the static def `metadata` becomes RunInput.Metadata
+// (trusted), and payload_mapping `run_metadata.*` targets become
+// RunInput.PayloadMetadata (untrusted). Fails on the pre-feature code, which
+// had no metadata fields and discarded run_metadata.* projection targets.
+func TestBuildRunInput_MetadataTrustedAndPayloadUntrusted(t *testing.T) {
+	w := config.Webhook{
+		Agent:    "reviewer",
+		Metadata: map[string]any{"policy": "strict", "skills": []any{"go"}},
+	}
+	proj := projectResult{Fields: map[string]string{
+		"goal":              "review the PR",
+		"run_metadata.repo": "acme/app",
+		"run_metadata.pr":   "42",
+	}}
+	in := buildRunInput(w, proj, map[string]bool{}, func(string) string { return "" }, nil)
+
+	if in.Metadata["policy"] != "strict" {
+		t.Errorf("static def metadata not carried as trusted Metadata: %v", in.Metadata)
+	}
+	if in.PayloadMetadata["repo"] != "acme/app" || in.PayloadMetadata["pr"] != "42" {
+		t.Errorf("run_metadata.* not routed to PayloadMetadata: %v", in.PayloadMetadata)
+	}
+}
+
+// TestBuildRunInput_UserCredentialsPrecedence pins env < fork-time < payload.
+func TestBuildRunInput_UserCredentialsPrecedence(t *testing.T) {
+	w := config.Webhook{
+		Agent:                  "x",
+		UserCredentialsFromEnv: map[string]string{"slack": "SLACK_ENV", "jobs": "JOBS_ENV"},
+		UserCredentials:        map[string]string{"slack": "fork-slack", "tg": "fork-tg"},
+	}
+	proj := projectResult{Fields: map[string]string{
+		"user_credentials.slack": "payload-slack",
+	}}
+	env := map[string]bool{"SLACK_ENV": true, "JOBS_ENV": true}
+	getenv := func(k string) string {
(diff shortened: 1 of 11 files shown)


denn-gubsky:
## Self-review pass (high-effort, whole feature: #356 + #357)

Reviewed the combined feature diff. **Applied to this PR (672ace1):**
- **#4** — `run_metadata.*` projection now skips empty values (an absent optional body path no longer surfaces as a phantom `{name:""}` in `PayloadMetadata`).
- **#9** — folded the two near-clone prefix-scan loops (`user_credentials.*` + `run_metadata.*`) into a shared `collectPrefixed(fields, prefix)`, so the namespaces can't drift in scan/skip semantics. (Fixes #4 for both.)
- Fail-before test `TestBuildRunInput_EmptyProjectedValuesSkipped`.

**Refuted on inspection — #10** (the `idx`-after-system logic in `injectMetadataSegments`): NOT dead complexity. `splitSegments` (loop.go:1230) appends `system` blocks in slice order and providers render them in order, so the idx logic is load-bearing — it keeps the agent's own system prompt leading and the metadata block second. Removing it would push caller metadata ahead of the agent identity prompt. Left as-is.

### Deferred to a follow-up PR (these touch #356-merged files: server.go / codejs / connector)
- **#1 (medium):** an LLM run's metadata block is injected into live Segments only, never persisted — a `/v1/sessions/{id}/messages` continuation that doesn't re-send metadata replays prior history without it (and it's system-role, which `replayTranscript` skips anyway). Needs a decision: persist it, or document per-run re-supply as the model.
- **#2 (medium):** `connector.SpawnRunRequest` has no `Metadata` field, so gRPC / LoomCycle-MCP `spawn_run` / HTTP-connector can't supply metadata (delivery via `RunOnce` is shared; only the supply field is missing). `ParentContext` is threaded there — `Metadata` is the parallel omission. Clean parity fix.
- **#3 (low/robustness):** `injectMetadataSegments` hardcodes `providerID == "code-js"`; a future structured-input provider would get metadata both as `input.metadata` AND as a shadowing untrusted user segment. Gate on a capability, not the literal string.

### Noted, not a code change
- **#7:** wire `metadata` is system-trusted-text reachable by any `runs:create` token — **pre-existing** (arbitrary system-role `segments` are already accepted), not introduced here. Worth a doc line in the multi-tenant security model.
- Lower-severity: payload_metadata passed by reference in `buildInput` (aliasing asymmetry vs the copied Metadata); MarshalIndent-error silently drops the segment (unreachable for wire-decoded maps).

PR #397, opened 2026-06-08
docs: clarify file-tool CWD + document process-resource metrics (F13/F19)

denn-gubsky:
## Why

Class E of the experiments change-plan — the documentation sweep. Several items
were **already correct** on `main` (fixed in releases after the v0.22.0 findings
snapshot), so this PR is the two genuine gaps:

- **F3 / E1** (help topic "nine vs twelve" + basic-group double-count) — already
  fixed: `loomcycle.md` says "twelve higher-level tools" with the basic group
  (`Read/Write/Edit/HTTP/WebFetch/WebSearch/Bash`) correctly *excluding*
  Grep/Glob/NotebookEdit. No change.
- **F4 / E2** (`Agent` vs `spawn_run` mapping) — already documented: the
  `subagents` help topic has an "`Agent` (in-loop) vs `spawn_run` (MCP surface)"
  section. No change.

## What

- **F13 / E4** — `docs/TOOLS.md` sandbox semantics now spell out that file-tool
  **relative paths resolve against the loomcycle process's CWD**, *not* against
  `LOOMCYCLE_WRITE_ROOT` / `READ_ROOT` (those are the absolute jail the resolved
  path must fall inside, not a join base). Launching loomcycle from a different
  directory than the jail root offsets relative `Write` paths from `Bash`'s
  `LOOMCYCLE_BASH_CWD` — the "agent wrote to `./foo` while Bash ran in `./work/`"
  footgun. Recommends launching from the jail root (CWD == root ==
  `LOOMCYCLE_BASH_CWD`) or using absolute paths.
- **F19 / E5** — the `observability` help topic gains a **Process-resource
  metrics** section. The OTEL-only topic never mentioned the `/v1/_metrics`
  sampler; this documents that it's off by default
  (`LOOMCYCLE_METRICS_ENABLED=1`), that **system-wide CPU/memory stay OFF even
  when metrics are on** until `LOOMCYCLE_METRICS_COLLECT_SYSTEM=1` (so the
  `system_*` columns read NULL and co-tenant host pressure is invisible — the
  exact F19 incident), and that the sampler is **not a substitute** for an
  always-on external host monitor.

Docs-only. The plugin-side item (**F12 / E3** — `settings` env changes need a
full Claude Code restart, not `/reload-plugins`) is a separate PR in the
`claude-code-plugin-loomcycle` repo.

## What was tested

`internal/help` builds + tests green (the embedded `observability.md` still
loads). `docs/TOOLS.md` is a plain doc. No code change.

### DIFF
--- a/docs/TOOLS.md
+++ b/docs/TOOLS.md
@@ -55,6 +55,7 @@ Bash has additional warnings: it is **not a true sandbox** even when enabled. Ru
 
 Sandbox semantics (file tools):
 - Paths must resolve **inside** the sandbox root after full `EvalSymlinks` evaluation. Symlinks pointing outside the root are refused.
+- **Relative paths resolve against the loomcycle process's current working directory — NOT against `LOOMCYCLE_WRITE_ROOT` / `LOOMCYCLE_READ_ROOT`.** Those env vars are the *absolute jail* the fully-resolved path must fall inside; they are not the base a relative path is joined to. So if you launch loomcycle from a different directory than your jail root, a relative `Write` path (joined to the process CWD) and `Bash`'s `LOOMCYCLE_BASH_CWD` can point at different places — the offset that scattered files in the experiments harness (the agent wrote to `./foo` while `Bash` ran in `./work/`). **Launch loomcycle from the directory you set as the write/read root** (so process CWD == jail root == `LOOMCYCLE_BASH_CWD`), or have agents use absolute paths.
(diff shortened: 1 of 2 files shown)


denn-gubsky:
Reopen failed (branch was deleted, breaking the head-ref linkage). Restored as #398 with the same commit rebased onto current `main`.

PR #504, opened 2026-06-16
docs: model-aliases-first config/example + local-model guide; + runs-chip UI fix

denn-gubsky:
Docs + examples only, no code.

## What
**1. `loomcycle.example.yaml` — aliases-first.** The `models:` alias map moves to the **top**; tier candidate lists and per-agent pins now reference **bare aliases** (`- deepseek-flash`) instead of repeating `{provider, model}` pairs (v0.35.0 made bare-string aliases work in tier candidates). Comments present aliases as the recommended way to name models — edit one RHS to re-point every agent/tier. `defaults:` stays a concrete pin (it's not alias-expanded; noted inline).

**2. `docs/CONFIGURATION.md` — aliases-first + local-model guide.**
- §3 (Pattern 1) reworked aliases-first (models defined once, tiers reference them as bare aliases).
- §1 resolution-axes table reframes aliases as the **always-recommended** naming, not agent-`.md`-only.
- New **§6b. Local models (Ollama)** — the slow-local-model lessons from recent work: the two Ollama providers; the global `LOOMCYCLE_OLLAMA_LOCAL_NUM_CTX` knob (caps **and** reports; unset → Modelfile default; gauge reads `/api/ps`); header/idle timeouts; the heartbeat-during-call behavior; compaction tuned for prefill cost (compact early, small `keep_last_n`); and the interactive knobs.
- §10 cross-references both example files.

**3. `loomcycle.local-interactive.example.yaml` (new).** A focused, ready-to-run config for steering interactive agents on local models: aliases, local-first tiers + cloud fallback, and a `coder` agent wiring `unbounded_iterations` + early/small-tail `compaction` + `interruption`, with the env knobs documented in the header.

## What was tested
- `loomcycle doctor --config <each>` → `[PASS] Config parses` for **both** example YAMLs (the new bare-alias tiers + the alias-aware load-time validation).
- Self-reviewed the restructured example for coherence; relative doc links resolve to the two root example files.

### DIFF
--- a/cmd/loomcycle/embedded/loomcycle.example.yaml
+++ b/cmd/loomcycle/embedded/loomcycle.example.yaml
@@ -16,12 +16,46 @@
 # config setup from exfiltrating, say, ANTHROPIC_API_KEY into an MCP
 # server URL.
 
+# ─── Model aliases — name your models ONCE, here, FIRST ───────────────
+# A model alias maps a friendly name → (provider, real model id). Define
+# every model you use here, then reference it BY NAME everywhere else:
+# tier candidate lists (below), per-agent pins (`model: sonnet`), and the
+# per-agent `models:` override. This is the recommended way to name models
+# — editing one right-hand side re-points every agent/tier that uses the
+# alias (e.g. when Anthropic ships sonnet-5, change `sonnet:` once). A raw
+# `{provider, model}` pair still works anywhere an alias does; the alias
+# just avoids repeating the id. NOTE: `defaults:` (far below) is NOT
+# alias-expanded — pin it with a concrete provider+model.
+models:
+  # — Anthropic (key: ANTHROPIC_API_KEY) —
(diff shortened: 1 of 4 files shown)


denn-gubsky:
**Follow-up commit + long-running test result.**

- Added a small UI follow-up: the runs-page **`interactive` chip** stretched to fill its `1fr` grid column; `justify-self: start` makes it text-wide (no-op in the non-grid uses). See `7dbe897`.
- **Long-running test (validates the merged #502 + #503 + the compact-early config):** an interactive `code-reviewer` on `qwen3.6:max` (single local GPU) ran **standalone for 133 minutes** through several *successful* auto-compactions (e.g. `74.6k → 28.8k`) before hitting the indivisible-tool-chain compaction gap — now captured as **RFC AD** (doc-internal). Pre-robustness-pass the same run died in minutes.

PR #733, opened 2026-07-15
feat(providers): P3 — delete the hardcoded built-in provider floor

denn-gubsky:
## Why

RFC BF **Phase 3 — "delete the built-ins."** P2a (#729) already flipped resolution fully config-driven — `newProviderResolver` builds only from `cfg.Providers` via the driver registry, and the boot probe iterates `cfg.Providers` — so the resolver's hardcoded fields, env-gated `New()` block, and `Get()` switch (the RFC's literal P3 targets, written pre-P2a) were **already gone**. What remained were the P2a **scaffolding enumerations** that still hardcoded the built-in set. P3 removes them so the embedded default-providers layer is the *sole* source of the built-ins.

## What

- **`cmd/loomcycle`** — deleted `expectedBuiltinProviderIDs` + the boot **divergence WARN** (a P2a Deliverable-5 transition scaffold comparing the merged config against a hardcoded 9-id list). There's no hardcoded expected set left to diverge from.
- **`internal/config`** — reduced `validProviderIDs` (the hardcoded 9-id validation **floor**) to `residualProviderIDs = {anthropic-oauth-dev}`. That's the one id valid to *reference* but deliberately not in the `providers:` map — RFC BF locked decision **#7** keeps it a thin residual (main-owned token `Refresher`; never a registry driver nor a `providers.default.yaml` entry). Every other built-in now validates only because the default-providers layer supplies it in `cfg.Providers`. Also replaced the stale hardcoded list in the `provider_priority` "unknown provider" message with the config's **actual** known set.
- **`internal/cli`** — `loadLayeredConfig` now **prepends the embedded default-providers layer** (honoring `LOOMCYCLE_NO_DEFAULT_PROVIDERS`), fixing a latent divergence: the CLI claimed to assemble "the SAME layered configuration the server builds" but omitted this base layer. This is what lets `validate`/`doctor`/`agents` accept a built-in-referencing config after the floor is gone. A **missing `--config` still errors** — the default-providers layer is infrastructure, not a user config source, so the explicit-path keep/skip decision keys off *user* layers only (`baseLayerCount`).

**End state:** the embedded default-providers layer is the sole source of the six real providers + `mock`/`mock-stable`/`code-js`, for **both** the server and the CLI. Under `LOOMCYCLE_NO_DEFAULT_PROVIDERS` the operator takes full control and must declare its own `providers:` (only the oauth-dev residual stays referenceable without one).

**Not in scope:** `defaultLibraryPriority` (a resolver default *ordering*, not a registration) and the `anthropic-oauth-dev` residual registration itself (kept per decision #7).

## What was tested

- **Migrated** the config/cli/cmd tests that built a built-in-referencing config *without* the default layer (they relied on the deleted floor): a `withDefaultProviders` helper prepends the embedded layer (mirroring the real load path); `validate()`-direct tests declare their built-ins inline; a CLI `drivers_test.go` registers the driver factories in the CLI test binary (as `cmd/loomcycle` and the config test binary do). One CLI regression found + fixed along the way: a missing `--config` must still error despite the always-present base layer.
- **New guards** (fail-before verified):
  - `TestValidate_NoBuiltinFloor_AfterP3` — a providers-less `provider_priority: [anthropic]` is rejected. Fail-before: restore `anthropic` to the floor → it passes.
  - `TestLoadLayeredConfig_PrependsDefaultProviders` — a providers-less `--config` referencing `anthropic` loads via the CLI; `NO_DEFAULT_PROVIDERS=1` makes it fail. Fail-before: drop the prepend → the first load rejects.
  - `TestValidate_OAuthDevResidualValidWithoutDeclaration`, `TestValidate_BuiltinValidViaDefaultProvidersLayer`.
- `go test -race ./...`, `go vet`, `gofmt -l`, `make build-all` — all clean.
- **Manual (real binary):** `loomcycle validate` on a providers-less config referencing built-ins → **OK** (probe → `provider=anthropic`); with `LOOMCYCLE_NO_DEFAULT_PROVIDERS=1` → fails `provider_priority[0]: unknown provider "anthropic" (declare it under providers:; known: [anthropic-oauth-dev])`; boot logs `providers: 9 configured, 2 enabled (anthropic, ollama-local)` with **no** divergence WARN.

### DIFF
--- a/cmd/loomcycle/main.go
+++ b/cmd/loomcycle/main.go
@@ -3014,15 +3014,6 @@ func buildEmbedder(cfg *config.Config) (providers.Embedder, error) {
 	})
 }
 
-// expectedBuiltinProviderIDs is the set the embedded default-providers layer
-// declares. Used only by the boot divergence WARN (RFC BF P2a Deliverable 5) —
-// resolution itself is fully config-driven, this just catches a default-layer
-// typo or a surprise LOOMCYCLE_NO_DEFAULT_PROVIDERS run.
-var expectedBuiltinProviderIDs = []string{
-	"anthropic", "code-js", "deepseek", "gemini", "mock", "mock-stable",
-	"ollama", "ollama-local", "openai",
-}
-
 // newProviderResolver builds every enabled provider from cfg.Providers via the
 // RFC BF driver registry (config-driven flip of the pre-P2a hardcoded per-provider
 // construction). It is byte-identical when no operator `providers:` block is
@@ -3272,9 +3263,10 @@ func providerDisabledReason(id string, pc config.ProviderConfig) string {
 	}
 }
 
-// logProviderSummary logs the config-sourced provider inventory (RFC BF P2a
-// Deliverable 5) + a WARN if the merged config is missing a built-in the
-// default-providers layer should have supplied (catches a default-layer typo).
+// logProviderSummary logs the config-sourced provider inventory. RFC BF P3: the
+// built-ins are now sourced SOLELY from the embedded default-providers layer, so
+// the hardcoded-vs-config divergence WARN (a P2a transition scaffold) is gone —
+// there is no hardcoded expected set left to diverge from.
 func logProviderSummary(pr *providerResolver, cfg *config.Config) {
 	enabled := make([]string, 0, len(pr.byID))
 	for id := range pr.byID {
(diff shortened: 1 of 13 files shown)


denn-gubsky:
Closing in favor of #734. On review, the decision was to KEEP the `validProviderIDs` validation floor rather than delete it: the hardcoded provider *construction/registration* that RFC BF P3 targeted was already removed in P2a, and what remains is a deliberate, tested validation floor (see `TestValidate_ThirdPartyProvider_AcceptedAsReference`) that keeps `validate()` a total function independent of the default-providers layer. #734 ships the same uncontested P3 cleanup — CLI default-providers-layer parity, dropping the P2a divergence-WARN scaffolding, and naming the actual known set in the `provider_priority` error — while retaining the floor. The floor-removal (Option A) in this PR is not the chosen direction, so closing to avoid two competing P3 PRs.

PR #750, opened 2026-07-17
feat(sandbox): builder-sidecar deploy guide + docker-socket image + gh in session image

denn-gubsky:
A detailed, phased guide to deploy the **isolated code-execution sandbox** (the builder sidecar) — TrueNAS-first, valid on any Docker host — plus the image variant that makes the recommended TrueNAS path actually work. **Docs/deploy only** — no Go/wire change, no tag needed.

## The wrinkle it solves

The sidecar needs a container engine to launch sessions. TrueNAS SCALE is Docker-based and has **no Sysbox**, so nested rootless podman is fragile there. The sidecar's engine already supports driving Docker (`SANDBOX_PODMAN_BIN=docker`) — it just needed a docker-CLI image. So the guide leads with the **host Docker socket** model: the sidecar drives the host engine and sessions run as **sibling** containers — no nested engine, no Sysbox, no `privileged`.

## What's here

- **`Dockerfile.docker`** — a small sidecar variant (`docker:cli` base + the static Go binary) for the socket model. The podman/stable `Dockerfile` stays for the nested model.
- **`INSTALL.md`** — the walkthrough:
  1. build + push the two images (sidecar + session toolchain);
  2. a shared `SANDBOX_AUTH_TOKEN` / `LOOMCYCLE_SANDBOX_TOKEN`;
  3. add the `builder-sidecar` service to the **same** loomcycle app compose (separate TrueNAS custom apps don't share a Docker network, so loomcycle must be able to resolve `builder-sidecar:9000`), mounting the host socket;
  4. enable the `sandbox` bundle on loomcycle;
  5. verify end-to-end (incl. a host-side `docker ps --filter label=loomcycle.managed=1` check).
  Plus the **nested-podman (privileged)** and **dedicated-VM** alternatives, the security posture, and a troubleshooting table.
- **`README.md`** — Deploy section now describes the two engine models and points at `INSTALL.md`.

## Notes / decisions

- **Sidecar as a service in the loomcycle compose**, not a standalone TrueNAS app — that's the only way the two share a network on TrueNAS.
- **The socket mount is a real trust grant** (≈ host root) — documented as the sidecar's designated privileged role; a compromised loomcycle still only reaches the constrained `sandbox_*` API. The VM model is offered for a hard boundary.
- One nicety worth calling out: **compiled binaries (`./a.out`) run fine in the sandbox** (a real container shell), unlike the toolbox's Bashbox fallback.

## Tested

- The sidecar module still builds (`go build ./...`); no Go change (it already speaks `docker`).
- All four `INSTALL.md` compose snippets parse as YAML.
- ⚠️ The images aren't built/run here (no local docker daemon) — build + push happens on a Docker host per the guide. I can add goreleaser targets to **publish** `loomcycle-builder-docker` + `loomcycle-sandbox-session` (so you'd `pull` instead of build) as a follow-up if you want.

### DIFF
--- a/deploy/builder/Dockerfile.docker
+++ b/deploy/builder/Dockerfile.docker
@@ -1,0 +1,42 @@
+# loomcycle-builder — HOST-DOCKER-SOCKET variant of the sandbox sidecar.
+#
+# Instead of running its own container engine (the podman-nested model in the main
+# Dockerfile), this variant drives the HOST's Docker engine over a mounted socket,
+# launching each sandbox session as a SIBLING container on the host. That's the
+# best fit for a Docker-based host that has no Sysbox — notably TrueNAS SCALE — and
+# it avoids nested-engine fragility + the `privileged` flag the nested model needs.
+#
+# Run it with:
+#   - the host Docker socket mounted:  /var/run/docker.sock:/var/run/docker.sock
+#   - SANDBOX_PODMAN_BIN=docker        (the engine shells `docker` instead of podman)
+# See ./INSTALL.md for the full TrueNAS walkthrough.
+#
+# ⚠️  Mounting the docker socket lets THIS sidecar control the host engine (≈ host
+# root). That is the sidecar's designated privileged role — a compromised loomcycle
+# still only reaches the constrained sandbox_* API, never the socket. Keep the
+# sidecar in-network (no host port), bearer-authed, single-tenant/trusted.
+#
+# Build:
+#   docker build -t denngubsky/loomcycle-builder-docker:dev \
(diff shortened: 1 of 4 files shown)


denn-gubsky:
Added a commit: the sandbox **session image** now installs `gh` (from the official apt repo). It already had the rest of the dev toolchain — git, curl, rsync, python3/pip3, go/gofmt, node/npm/npx, gcc/g++/cc/c++, clang/clang++, make/cmake/pkg-config (+ cargo/rustc) — so `gh` was the only gap vs. the requested list. The toolchain lives in the session image (what the sandbox runs code in), not the sidecar image (which only orchestrates).

PR #754, opened 2026-07-17
feat(mcp): forward ${run.root_run_id}/${run.tenant_id} to MCP servers (RFC BI P2b)

denn-gubsky:
The **loomcycle-core half of RFC BI P2b** — forwarding the run's non-secret identifiers to a downstream MCP server so the sandbox sidecar can attribute sessions to a run tree. (The sidecar half — `sandbox_touch`, `sandbox_close_run`, header-tagging, the run-liveness poll — is a follow-up PR on `deploy/builder`.)

## What

- `internal/tools/mcp/http` — new **`substituteRunIDs`** adds `${run.root_run_id}` and `${run.tenant_id}` substitution to outbound MCP header values, alongside the existing `${run.user_bearer}` / `${run.credentials.<n>}` / `$cred:`. These are **non-secret identifiers**, so an unresolved token substitutes to **empty** (not a dropped header). Disjoint from the bearer/credential regexes; runs after them; never drops.
- The **`sandbox` bundle** forwards `X-Loom-Root-Run: ${run.root_run_id}` + `X-Loom-Tenant: ${run.tenant_id}` on `mcp_servers.sandbox` (source ⇄ embed mirror identical).

## Why

P2b's `sandbox_close_run` + run-liveness GC need each session tagged with its owning run tree, and it must be **attested** (loomcycle-filled, not agent-supplied) — this is the substitution that delivers it. `X-Loom-Tenant` is forwarded now too, setting up the later attested per-tenant binding.

## Tested

- `TestSubstituteRunIDs` — resolves both, empty-on-missing (no dropped header), no-token passthrough, and leaves `${run.user_bearer}` untouched.
- Full `go test ./...` clean; the sandbox bundle still loads + validates (`TestEmbedded_SandboxBundleValidates`); gofmt clean.

Additive; no wire/schema change; adapters unchanged. Design: RFC BI P2 (`/loomcycle/rfcs/sandboxed-code-execution-p2`).

### DIFF
--- a/bundles/sandbox/loomcycle.yaml
+++ b/bundles/sandbox/loomcycle.yaml
@@ -159,3 +159,7 @@ mcp_servers:
       # Per-run bearer when present, else the shared LOOMCYCLE_SANDBOX_TOKEN secret
       # (must equal the sidecar's SANDBOX_AUTH_TOKEN).
       Authorization: "Bearer ${run.user_bearer:-${LOOMCYCLE_SANDBOX_TOKEN}}"
+      # Non-secret run identifiers the sidecar tags each session with, for
+      # run-liveness GC + sandbox_close_run (empty when the run carries none).
+      X-Loom-Root-Run: "${run.root_run_id}"
+      X-Loom-Tenant: "${run.tenant_id}"
--- a/bundles/sandbox/loomcycle.yaml
+++ b/bundles/sandbox/loomcycle.yaml
@@ -24,7 +24,7 @@ agents:
   dev/sandbox:
     tier: middle
     effort: medium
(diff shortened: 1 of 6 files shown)


denn-gubsky:
Added commit `fcbe0de`: grant `sandbox_touch` + `sandbox_close_run` to the `dev/sandbox` agent + skill (both bundle copies + a regression assertion in `TestEmbedded_SandboxBundleValidates`). Without this the sidecar advertises the two P2b tools but the agent can't reach them.

**P2b pair** — this is the loomcycle-core + bundle half; the sidecar half (the two new tools + attested run tagging) is **#755**. Independent + graceful in either merge order: until the sidecar (#755) is deployed the new tools are registered-but-unreachable (MCP pool retries lazily), and until this PR sends the header the sidecar just leaves sessions untagged.

PR #755, opened 2026-07-17
feat(sandbox): sandbox_touch + sandbox_close_run + run tagging (RFC BI P2b)

denn-gubsky:
The **sidecar half of RFC BI P2b** (long-session lifecycle). Companion to **#754** (the loomcycle-core `${run.root_run_id}` substitution + the bundle header). Sidecar-only (`deploy/builder`).

## What

- **Attested run tagging.** The MCP handler reads the `X-Loom-Root-Run` (+ `X-Loom-Tenant`) headers loomcycle forwards and carries them in a `caller` struct alongside the bearer-derived principal; `sandbox_open` tags the session's `RootRun` from the **header, never a tool arg** (unforgeable).
- **`sandbox_touch(session_id)`** — reset the idle timer; a keepalive so a driver can hold a container across a gap between commands without the idle TTL reaping it.
- **`sandbox_close_run(root_run_id)`** — bulk-close every session the caller owns for a run tree (`Store.RemoveByRun`): principal-scoped, non-empty-guarded (never sweeps untagged sessions), idempotent.

## Deferred (documented)

**Auto-reap on run-end** — a run-liveness poll of loomcycle *or* a loomcycle run-end push seam. It needs the `GET /v1/runs/{id}` status contract nailed down + a loomcycle poll token + a live deployment to verify, none of which are testable here. The **absolute TTL is the backstop**, and with durable workspaces (P2a) a reaped container loses no work — so this is an optimization, cleanly scoped as a follow-up.

## Graceful without #754

Until #754 lands (loomcycle sending the header), the header is simply absent → sessions are untagged: `sandbox_touch` still works; `sandbox_close_run` finds nothing; the TTL still reaps. No breakage.

## Tested

`go test -race ./...`:
- `sandbox_touch` advances `LastUsed`; unknown-session + foreign-principal refused.
- `sandbox_close_run` closes exactly the run's sessions, leaves run-B + untagged alone, refuses empty `root_run_id`, and a foreign principal closes 0.
- Header→session tagging verified through the real HTTP handler (open with `X-Loom-Root-Run`, then `close_run` finds it).
- The `Call(principal)` → `Call(caller)` refactor threaded through all tests; gofmt/vet clean; root module unaffected.

Design: RFC BI P2 (`/loomcycle/rfcs/sandboxed-code-execution-p2`).

### DIFF
--- a/deploy/builder/README.md
+++ b/deploy/builder/README.md
@@ -33,7 +33,14 @@ loomcycle-builder (this sidecar)  ──►  per-session container
 | `sandbox_exec` | Run a command in a session (`/work`); returns combined output + exit code. |
 | `sandbox_write` / `sandbox_read` | Put files in / pull artifacts out (paths relative to `/work`; `base64` for binary). |
 | `sandbox_close` | Destroy a session (idempotent). |
+| `sandbox_touch` | Reset a session's idle timer — a keepalive to hold a container across a gap between commands. |
+| `sandbox_close_run` | Close ALL your sessions for a given `root_run_id` (bulk teardown of a run tree). |
 | `sandbox_list` | List your own sessions. |
+
+Sessions are tagged with their owning loomcycle **run** via the attested
+`X-Loom-Root-Run` header loomcycle forwards (`${run.root_run_id}`) — that's what
+`sandbox_close_run` matches on. The tag is set from the request header, never a
+tool argument, and it's principal-scoped (a caller only ever closes its own).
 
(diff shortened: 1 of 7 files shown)


denn-gubsky:
**P2b pair** — sidecar half. The loomcycle-core companion (the `${run.root_run_id}`/`${run.tenant_id}` header substitution + the bundle's `X-Loom-Root-Run` header + the `dev/sandbox` agent tool grants) is **#754**. Independent + graceful in either merge order (see #754).

PR #798, opened 2026-07-23
feat(retention): RFC BM P3 — retired-agent memory-scope reclamation

denn-gubsky:
## Why
RFC BM P1 (#796) freed retired **def versions**; P2 (#797) freed aged **chats**. But a retired agent's accumulated per-scope **data** was still immortal — its SQL databases + document structure (SQL Memory), document bodies + learned facts (base memory k/v), and Path names (dirents) all outlive the agent forever. P3 reclaims a fully-retired agent's data, opt-in, closing the RFC BM memory family and completing the RFC.

## What
A `mem` sub-sweep in `internal/retention`. An agent whose **every** def version is retired (`AgentDefListNames` `LiveVersionCount==0`) and whose latest version is older than `MemMaxAge` is reclaimed in **two passes with different tenant-keying** — because base memory and the other facets key differently:

- **Pass 1 — per `(tenant, name)`, tenant-qualified (safe):** drop the agent's **SQL-Memory scope** (new `sqlmem.Manager.DropScope`, wrapping the existing `dropScopePG` / sqlite fenced-remove; rejects the run scope + empty key) and its **dirents** (`DirentDeleteUnder` at the scope root).
- **Pass 2 — per `name`, GLOBALLY DEAD only:** drop the agent's **base-memory k/v** (new `store.MemoryDeleteScope`; embeddings cascade via FK). Base memory is keyed by the **bare agent name** (the `memory` table has no tenant column), so it's shared by same-named agents across tenants — dropping it is safe **only when the name is retired in every tenant**. A name live in any tenant keeps its base memory. **This is the load-bearing cross-tenant safety invariant.**

**export-then-delete:** `export+prune` dumps each facet to `<ExportDir>/agents/<day>/{sqlmem,dirents,agent-memory}/<tenant>__<name>.json` before dropping (`fsSafe` filenames for `/`-grouped agent names). SQL-Memory export/drop is gated on `ListScopes` membership so `ExportScope` never provisions an empty scope; a failed export or scope drop **aborts that agent's dirent drop** so names are never orphaned from a still-present scope. `memEligible` is shared by the sweep and the report so the count can't drift.

**Config:** `LOOMCYCLE_RETENTION_MEM_MODE` / `_MEM_MAX_AGE_MS` (opt-in, default off). `main.go` hoists the `sqlMemMgr` declaration so the sweeper can reach it (nil-safe: base memory + dirents still reclaim when SQL Memory is disabled). `GET /v1/_retention` gains `mem_mode`/`mem_max_age` (tenant-readable) + `purgeable["mem"]` + a `sqlmem_gc` block surfacing the separate always-on SQL-Memory GC (admin-only).

**Scope decision:** document age-retention was **dropped** from P3 (operator-confirmed) — the loomcycle document store now holds every RFC and guide, so an age-based purge could delete them, and documents have no cross-scope age-enumeration surface. Explicit `delete_document` remains.

## Tested
- Store contract `MemoryDeleteScope` (sqlite + **real Postgres**) — deletes a whole scope, leaves siblings + same-id other-scope alone, idempotent.
- sqlmem `TestDropScope_SQLite` + `TestDropScope_Postgres` — drops one durable scope idempotently, leaves siblings, rejects run scope + empty key.
- retention `TestSweeper_Mem{ReclaimsRetiredAgent, BaseMemoryOnlyDroppedWhenGloballyDead, ExportThenDelete, ExportFailureSkipsDrop, DryRunCounts}`.
- **Cross-tenant safety is fail-before verified:** dropping the globally-dead gate makes `TestSweeper_MemBaseMemoryOnlyDroppedWhenGloballyDead` fail (base memory deleted while a live tenant still uses it).
- `gofmt` + `go vet` clean; full affected suites green on sqlite + Postgres.

> ⚠️ Pre-existing unrelated failure: `TestStoreContract/ChannelPurge` fails identically on **clean HEAD** (verified via a worktree) — not touched by this change. Flagging separately.

### DIFF
--- a/cmd/loomcycle/main.go
+++ b/cmd/loomcycle/main.go
@@ -2377,8 +2377,13 @@ func main() {
 	// sweeper owns aged-chat pruning — suppress the usage sweeper's legacy
 	// aged-session archiver so a session is never cascade-deleted by both. The
 	// usage-detail rollup is unaffected.
-	retentionOwnsChats := cfg.Env.RetentionEnabled &&
-		cfg.Env.RetentionChatsMode != "" && cfg.Env.RetentionChatsMode != "off"
+	// Only suppress the legacy usage archiver when the retention chats sweep will
+	// ACTUALLY run — mirror retention's chatsPruneEnabled gate (export+prune needs
+	// an ExportDir). Otherwise a misconfigured export+prune-without-dir would
+	// disable BOTH paths, silently leaving aged chats unpruned.
+	retentionChatsActive := cfg.Env.RetentionChatsMode == "prune" ||
+		(cfg.Env.RetentionChatsMode == "export+prune" && cfg.Env.RetentionExportDir != "")
+	retentionOwnsChats := cfg.Env.RetentionEnabled && retentionChatsActive
 	if cfg.Env.UsageSweeperEnabled && storeIface != nil {
 		uCfg := usage.Config{
 			Interval:        cfg.Env.UsageSweepInterval,
--- a/cmd/loomcycle/main.go
+++ b/cmd/loomcycle/main.go
@@ -1681,6 +1681,9 @@ func main() {
 	// SEPARATE from the main store — an agent's arbitrary SQL can only ever
 	// reach its own scope file. Wired into the Memory tool (sql_query/sql_exec)
 	// and the server (run-scope drop at top-level run completion).
+	// Declared here (not inside the block) so the RFC BM retention sweeper below
+	// can reclaim a retired agent's SQL-Memory scope; nil when SQL Memory is off.
+	var sqlMemMgr *sqlmem.Manager
 	if cfg.Storage.SqlMemEnabled {
 		sqlMemRoot := cfg.Storage.SqlMemRoot
 		if sqlMemRoot == "" {
@@ -1705,7 +1708,6 @@ func main() {
 		// sqlite deploy gets file-per-scope. The aux DSN is required on
 		// postgres and ignored on sqlite.
 		var (
-			sqlMemMgr  *sqlmem.Manager
 			smErr      error
 			sqlMemTier string
 		)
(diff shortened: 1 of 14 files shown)


denn-gubsky:
## Self-review of the full RFC BM feature (P1+P2+P3)

Ran a high-effort code review over the entire retention feature line. **No data-loss bugs** — every destructive path is correctly scoped (def purge leaf-only + active-excluded + tenant-partitioned; chats pinned-exempt + session-cascade; P3 memory tenant-safe), and all defects fail toward *under*-reclamation, never over-deletion.

Found and **fixed two real correctness bugs** in P3 (pushed as `89c2d33e`):

1. **Pass-2 base-memory reclamation gated on `MemoryListScopeIDs` (hardcoded `ORDER BY updated_at DESC LIMIT 200`).** A retired agent's memory is stale, so with >200 agent memory scopes it sorts out of that window and was never reclaimed — pass 2 skipped exactly its targets. Removed the cap dependency (direct `MemoryDeleteScope` + a limit-1 existence probe for the count). Regression `TestSweeper_MemBaseMemoryReclaimedBeyondScopeIDsCap` (fail-before verified).
2. **`sweepOnce` ran the defs purge before the mem reclaim.** With `keep_last_n=0` the same tick could purge an agent's last def version, orphaning its SQL-Memory scope + base memory + dirents. Reordered mem-before-defs. Regression `TestSweeper_MemRunsBeforeDefsPurge` (fail-before verified).

Plus two low-severity fixes: export-truncation now skips the delete (export-then-delete preserved for a >1M-key agent), and `retentionOwnsChats` now only disables the legacy archiver when the retention chats sweep will actually run.

Three PLAUSIBLE low-severity findings left as documented (an `fsSafe` filename collision for names differing only by `/`↔`_`; the mem sweep re-scanning already-reclaimed retired agents each tick; a per-candidate `DirentListUnder` on the admin report). Happy to address if you'd like.

gofmt + vet clean; retention + config suites green on sqlite; fixes touch no SQL so the Postgres store paths are unchanged.

PR #875, opened 2026-07-29
feat(memory): tenant scope for Memory, SQL Memory and Documents (RFC BL P4b)

denn-gubsky:
The prerequisite the rest of P4 sits on: a **tenant scope** — one keyspace per tenant, shared by every user and agent inside it — alongside the existing per-agent and per-user scopes.

## The architecture pass moved most of the work

**sqlmem needed almost no code.** It's already fully generic over the scope string: nothing in the package validates it, `pgScopeNames` hashes `sha256(tenant \x1f scope \x1f scope_id)` for any scope, the sqlite backend joins `root/<tenant>/<scope>/<id>.db`, and `scope_registry` records it verbatim. A tenant scope flows through the existing provisioning path untouched.

**The real gap was that k/v Memory had no tenant scope at all.** RFC AK splits a Document across two planes — structure in SQL Memory, chunk **bodies** in Memory at `doc.chunk:<uuid>` — so a tenant-scoped Document needs a tenant-scoped Memory partition for its bodies. That makes this three planes, not two. No DDL: `memory.scope` is a plain text column with no CHECK, and the tenant axis arrived with migration 0059.

**The two planes need different scope ids for the same logical scope**, and the asymmetry is load-bearing:

| plane | scope_id | why |
|---|---|---|
| sqlmem | the tenant | `pgScopeNames` **rejects** an empty component, and the value becomes half of a schema + LOGIN role name |
| Memory | `""` | the `tenant_id` column already carries the identity; a second copy leaves the retention fan-out ambiguous about which is authoritative. Matches global's convention and the Path tree's tenant dirent |

Both sites say so with the reason, because the next reader will otherwise "fix" one of them.

## Authority — the grant is the gate, deliberately

A tenant scope is shared-**write**: anything an agent stores is read by every other agent and user in the tenant. That's the poisoning surface the memory research flags. It stays default-deny through `memory_scopes:` / `sql_scopes:`, the same per-agent allowlists that gate the other scopes.

**No second mechanism was added.** The RFC's "agents propose, operator approves" is about the *ontology* (a P4c concern); a parallel gate here would leave two places to get wrong instead of one. `validMemoryScopes`'s own comment had already anticipated the value: *"future versions may add session / tenant"*.

## Nothing enumerating scopes needed changing — verified, not assumed

Snapshot capture reads every row and passes the scope through as a string; retention targets the agent scope specifically (a tenant scope has no retirement event to sweep on); the quota and GC paths take the scope triple generically.

## Two tests encoded the pre-P4b state

Replaced, not deleted. `TestMemoryScopesValidation` used `tenant` as its example of an **invalid** scope, so it now rejects a genuinely unknown one (`session`) — a validation test whose bad input later becomes valid stops testing anything, silently. `TestDocument_TenantScopeRefused` becomes `TestDocument_TenantScopeRoundTrips`.

## What was tested — both tiers

**Postgres, against a real aux DB** (the plan named per-tenant role creation as the sharp risk precisely because it fails at provisioning time rather than review time, so it's asserted):

- the schema **and** the LOGIN role are created
- the scope lands in `scope_registry`, so snapshot capture cannot skip it
- `DropScope` reclaims it — which only works because the tenant scope's id is non-empty

**sqlite**: provisioning, cross-scope and cross-tenant isolation, and that `{t,tenant,t}` does not alias `{t,user,t}` — they differ only in the hashed name's scope segment, which is exactly what a naming refactor breaks silently.

**Memory level**: shared-across-users (the scope's purpose), isolated-across-tenants (what makes it safe to grant), default-denied-without-the-grant (the gate holding), and no-collision-with-global — both use `scope_id=""`, separated only by the scope column and the tenant partition.

**Document level**: a tenant document round-trips through *both* planes including its chunk body, and is not reachable through the user scope. A mismatch between the planes would present as a document whose structure exists and whose text is empty — a failure this store has already had once.

`go test ./...` green on both tiers, including the 262s Postgres store contract. `gofmt` clean.

### DIFF
--- a/internal/api/mcp/context.go
+++ b/internal/api/mcp/context.go
@@ -112,9 +112,34 @@ func grantOperatorPolicies(ctx context.Context, agentName string, isAdmin bool) 
 	// actual run, and this plane has no run id. An operator writing a fact by hand
 	// is not a machine distilling one from a transcript, and the column has to stay
 	// a trustworthy filter for the latter. See builtin.provenanceForSet.
+	//
+	// `tenant` (RFC BL P4b) is granted here for BOTH admin and tenant principals.
+	// This plane IS the operator — a substrate:tenant token is the tenant operator,
+	// with full power inside its own tenant — and curating tenant-shared reference
+	// material is exactly what it is for. Confinement is the TenantID stamped from
+	// the principal, so a tenant token reaches only ITS tenant's tenant scope; the
+	// scope list is not what keeps tenants apart.
+	//
+	// It is deliberately WIDER than an ordinary agent's grant, because the trust is
+	// different in kind: an agent gets `tenant` only when an operator lists it in
+	// yaml, whereas this session is the operator, authenticated as themselves.
 	ctx = tools.WithMemoryPolicy(ctx, tools.MemoryPolicyValue{
-		AllowedScopes: []string{"agent", "user", "global"},
+		AllowedScopes: []string{"agent", "user", "tenant", "global"},
 		Consolidation: true,
+	})
+	// SQL Memory: never stamped before, which meant EVERY sql_* op over MCP failed
+	// default-deny with "this agent has no sql_scopes configured" — at any scope,
+	// since v1.2.0. Verified against a live deployment rather than inferred.
+	//
+	// Documents hid the gap: the Document tool talks to sqlmem directly instead of
+	// through the Memory tool's policy check, so document ops over MCP worked while
+	// the SQL ops beside them did not.
+	//
+	// `run` is omitted on purpose: an MCP-direct call has no run, so granting it
+	// would trade a clear "not configured" refusal for a confusing "requires an
+	// active run" one.
+	ctx = tools.WithSqlMemPolicy(ctx, tools.SqlMemPolicyValue{
+		AllowedScopes: []string{"agent", "user", "tenant"},
 	})
 	// Channel: open ACL ("*" matches every channel name).
 	ctx = tools.WithChannelPolicy(ctx, tools.ChannelPolicyValue{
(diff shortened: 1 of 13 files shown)


denn-gubsky:
## Update: tenant writes are now default-deny on every path, gated by both grants

Two follow-up commits after review. The original claim in this PR body — "the per-agent grant IS the gate" — was **true for the Memory tool and false for the Document tool**.

### The gap

| path | gated originally? |
|---|---|
| `Memory` — k/v tenant scope | yes (`contains(policy.AllowedScopes, requested)` before resolution) |
| `Memory` — SQL tenant scope | yes (same, against `sql_scopes`) |
| **`Document` — tenant scope** | **no. No grant check at all — it never had one.** |

An agent holding `tools: [Document]` could create tenant-scoped documents, readable and writable by every user and agent in the tenant, with the operator granting nothing.

**The evidence was already in this PR and I misread it as a pass:** `documentFixture` stamps no memory policy, and `TestDocument_TenantScopeRoundTrips` went green anyway. A test that passes without the grant is proof the grant is not required.

### The fix

`scope=tenant` now requires `tenant` in **both** `memory_scopes` and `sql_scopes`, because a document write touches both planes — structure in SQL Memory, chunk bodies in k/v Memory. Granting one and not the other would reach half a tenant store, and a half-written document isn't a partial failure: it's structure with no text, the exact shape this store has already been broken into once.

The refusal names which grants are *missing*, so an operator who has added one is told about the other rather than re-reading a message they already satisfied.

Fail-before verified: with the check in place, the previously-passing ungated round-trip fails with `scope=tenant is not granted to this agent`.

### Deliberately left alone

`agent` and `user` remain ungated on the Document path — pre-existing since the tool shipped. Requiring grants there would break every existing agent that holds `Document` without declaring scopes, which is a separate breaking change and must not arrive by accident inside a feature commit. `TestDocument_TenantGateLeavesOtherScopesAlone` pins that. It is now a recorded gap in the code rather than an unnoticed one, and worth its own PR.

### Tests

`TestDocument_TenantScopeRequiresBothGrants` covers all four combinations — neither / memory only / sql only / both — and asserts the refusal names the missing half in each case.

Full suite green on both tiers.

PR #946, opened 2026-08-04
fix(memory,document): both v1.46.0 upgrade sweeps were broken in practice

denn-gubsky:
Two bugs, both found by actually running the v1.46.0 upgrade sweeps against the reference deployment (154 documents, 3,143 chunks, 2 images). Neither was visible from the code or the unit tests.

## 1. The embedding backfill starved instead of finishing (`651f19a`)

Throughput decayed **189 / 179 / 169 / 160 / 147** embedded per 200-row window and was heading for zero with ~2,300 rows still unembedded.

`limit` bounded how many rows were **looked at**. A row with no body text — a document root, a section heading — can never gain an embedding, so it stays a candidate forever, and since the query is `ORDER BY key` those rows accumulate at the **front** of the window. The residue obeys `R' = R + p·(limit − R)`, whose fixed point is **R = limit**: eventually the window is entirely rows that cannot be embedded and the sweep does nothing — while its own notes promise "re-invoke until candidates reaches 0", a state it can never reach.

Widening the window didn't help either: the store did `if limit > 1000 { limit = 200 }`, so asking for 5000 silently returned a **smaller** page than asking for 1000. I hit that live while trying to work around the first bug. Now a clamp.

**Fix.** `MemoryEmbedListMissing` takes an `afterKey` keyset cursor, and the handler pages with it until it has *embedded* `limit` rows or the source is exhausted — so `limit` bounds work done rather than rows glanced at, and the sweep steps over what it cannot embed. The response reports `skipped_empty` and `more`, and the notes say to re-invoke while `more` is true.

Also: an admin token naming **no** tenant is refused with 400 `tenant_required`, mirroring erasure / directory / orphan-repair. Memory rows are keyed on the tenant, so omitting it resolved to the *default* tenant. Demonstrated live on the three-tenant deployment — the same request with and without `?tenant=` returned **200 candidates** and **0**. An explicit `?tenant=` still targets the default partition, or it would become unsweepable.

## 2. The describe pass truncated before it said anything (`f7e9119`)

The image sweep reported 2 candidates, 0 failures, and **both** as answered-empty. The model was fine — asked directly, qwen3.6 describes the image in 198 characters.

`describeMaxTokens` was 300, and qwen3.6 is a thinking model that emits its reasoning trace **first**. Reproduced straight against Ollama:

| request | content | thinking | done_reason |
|---|---|---|---|
| `num_predict=300`, think default | **0** | 1281 chars | **length** |
| `num_predict=300`, `think=false` | 448 | 0 | stop |

300 looked ample because the same prompt answers in 84 tokens with thinking off.

**Worse than a failed call.** The pass recorded it as answered-empty and stamped `described_at` — whose entire purpose is to separate "a model looked and found nothing" from "nothing has looked yet". So a code bug became a permanent fact about the data, and no re-run would ever revisit it. A turn that stopped at the ceiling is now a **failure**, left retryable, with the reason named.

The ceiling is 1500 — a ceiling, not a target, so a generous value costs nothing on a normal call. Deliberately **not** fixed by forcing `effort: low` (which the Ollama driver maps to `think:false`): that flag errors on a model which cannot reason, so it would break a non-thinking vision model like llava to accommodate a thinking one. A budget is model-agnostic.

## What was tested

- `TestBackfillEmbeddings_PagesPastRowsItCannotEmbed` — four unembeddable rows ahead of five embeddable ones at `limit=4`. The old single-window code embeds **0**; this embeds 4 and reports `more=true`.
- `TestBackfillEmbeddings_AdminMustNameATenant` — the refusal *and* the explicit-empty carve-out.
- Store contract, both tiers — the cursor is exclusive, past-the-end is empty rather than an error, and a limit over the cap clamps instead of falling back to the default page size.
- `TestDescribeImages_TruncatedAnswerIsRetryableNotEmpty` — `failed=1` / `empty=0`, the diagnostic names the ceiling, and a second pass still finds the image and describes it.
- `TestDescribeImages_BudgetLeavesRoomForAThinkingTrace` — pins the constant against the measured trace.

**Fail-before verified on both.** Reverting the paging loop reports `embedded=0`; disabling the truncation check reports `empty=1` and retires the image. Two existing backfill tests now name the tenant explicitly, since they relied on the ambiguous form.

`go test ./...` clean.

## Deployment note

The sweep itself completed — **2,814 chunk bodies embedded**, documents are semantically searchable now (verified: the query that returned `{"entries":[]}` before the sweep now returns the exact section at 0.62 similarity). 187 rows are the expected unembeddable residue.

After this deploys, the two images stamped by the buggy pass need clearing to become candidates again:

```sql
UPDATE chunk_assets SET described_at = NULL
 WHERE described_at IS NOT NULL AND coalesce(description,'') = '';
```

### DIFF
--- a/REVISIONS.md
+++ b/REVISIONS.md
@@ -44,7 +44,9 @@ POST /v1/_document/describe_images?scope=user&scope_id=<subject>&dry_run=false
 
 Both are resumable with no cursor: an embedded row or a described image drops out of its candidate set, so re-invoke until `candidates` reaches 0. Every chunk write now makes an embedder call — best-effort, so a cold or absent embedder logs and moves on rather than failing an author's write. Expect per-scope vector growth roughly equal to the chunk count, which brings the existing per-scope quota closer.
 
-Adapters: `@loomcycle/client` **1.46.0**, `loomcycle` (PyPI) **1.46.0** — the Python adapter had been pinned at 1.38.0, so this release publishes its erasure and directory methods. `@loomcycle/explorer` **0.6.0** (tag editing in the chunk editor).
+**The shared document viewer surfaces all of it.** `@loomcycle/explorer` gains read-only tag chips, a **Connections** panel with three lazily-fetched lists (backlinks — marking which edges came from `[[name]]` parsing — plus related and unlinked mentions, the three reads under `Promise.allSettled` so one failure never blanks the others, and a missing embedder showing a muted "not configured" note rather than an error), a **History** modal with the revision list, one revision's exact body and a unified diff, and a **Download .canvas** button beside Download .md. Board, kanban and spatial views stay a loomboard concern.
+
+Adapters: `@loomcycle/client` **1.46.0**, `loomcycle` (PyPI) **1.46.0** — the Python adapter had been pinned at 1.38.0, so this release publishes its erasure and directory methods — and `@loomcycle/explorer` **0.6.0**.
 
 ## What's in v1.45.0
 
--- a/REVISIONS.md
+++ b/REVISIONS.md
@@ -4,6 +4,29 @@ Per-version release notes from v0.4.0 onward. The current and immediately previo
 
 For the **public roadmap** (planned v0.8.16 through v1.0 work — Question tool, Pause / Resume / Snapshot, distribution, operator postures), see [`docs/PLAN.md`](docs/PLAN.md).
 
+## What's in v1.46.1
+
+**🩹 Both v1.46.0 upgrade sweeps were broken in practice.** Patch release — no new surface, no schema change. Found by running the sweeps against a real deployment (154 documents, 3,143 chunks, 2 images); neither fault was visible from the unit tests, because each needs scale or a real model to appear.
+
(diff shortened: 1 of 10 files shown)


denn-gubsky:
## CI fix + v1.46.1 notes (`c58686e`)

The postgres-tier failure was **my test's fault, not the code's**:

```
--- FAIL: TestStoreContractWithPgvector/MemoryEmbedListMissing
    contract.go:11143: cursor precondition: [doc.chunk:b]
```

My new cursor assertions called `set("doc.chunk:a", false)` to turn an already-embedded row back into a candidate — but that helper only writes the memory row and **returns early**; it never deletes an existing embedding. So `doc.chunk:a` stayed embedded, the precondition saw one row instead of two, and the test failed for its own reason rather than for the code's.

It passed locally because the whole function is gated on `vectorRefusalCheck`, which **skips on sqlite** — so those assertions had never executed anywhere until the postgres tier ran them. Exactly the "asserted against a fixture I hadn't read" failure mode.

**Fix:** use a fresh key (`doc.chunk:d`) and move the block to the *end* of the function, so no earlier state is disturbed and the pre-existing unprefixed-sweep assertion keeps its expected two rows. No production code changed.

Verified against `pgvector/pgvector:pg16` locally — the same image CI runs, since my usual local container is plain `postgres:16-alpine` and silently skips the vector suite.

Also adds the **v1.46.1** release notes so the patch can be tagged from one merge. No adapter changes in this PR, so their versions stay at 1.46.0 / 0.6.0 and their publishes will skip-but-green, which is correct.

PR #1040, opened 2026-08-17
feat(memory-view): the two embedding-maintenance ops, and 0.2.0 to publish them

denn-gubsky:
RFC CF phase 5's other half — unblocked by `@loomcycle/client@1.55.0` reaching npm — plus the version bump the package needs to publish any of the last month's work.

## The ops

`backfill_embeddings` and `purge_stale_embeddings` join the reembed flow they belong beside: all three are per-scope, which is what settled where they live, and all three are triggered from the same embed-stats badge.

Both follow reembed's **plan-then-commit** shape. `dry_run` defaults true server-side and the console always states it explicitly, so a plan can never become a commit through an omitted flag. The purge additionally confirms, because it deletes.

## The banner exists for two numbers a bare count misrepresents

- **backfill's `skipped_empty`** — rows with *no text to embed* (a document root, a section heading) remain candidates permanently. Unsaid, an operator watches `candidates` stop falling with `embedded` at 0 and no stated reason. **`more`** reports the limit was reached with work outstanding — the honest replacement for "run it until candidates hits 0", a target unembeddable rows make unreachable.
- **purge's `truncated`** — the scan stopped at the limit, so a zero `stale` does **not** mean the scope is clean. This op deletes, and *"we looked at some of it and found nothing"* is a different statement from *"there is nothing"*.

## A gap in the adapter, found by trying to use it

`@loomcycle/client@1.55.0` ships both methods but **does not re-export their response interfaces** from its entry point — they're defined in its types module and missing from the `export type {…}` list, so a consumer cannot name them. Fixed here for the next release.

Meanwhile the package **derives** those types from the client's own method signatures rather than declaring copies. That's drift-free (the type *is* the client's) and meant this work didn't have to wait on a second release to use methods that already exist.

## Version 0.2.0

The peer dependency moves to `^1.55.0` — the package now needs a client that has these methods — and the lockfile moves with it.

This bump is the prerequisite for a `memory-view-v0.2.0` tag. The package has been **13 files and ~866 lines behind its published 0.1.0** since the facts/verdict surface moved into it, so nothing consuming the package (rather than this repo's own web app) has any of the spans, verdicts, coverage strip, `remember`, or the vouching notice.

## Tested

42 package tests, `tsc --noEmit` clean, `npm run build:lib` clean, `make build-ui` and `go test ./...` clean. The dry-run pass-through is verified fail-before.

### DIFF
--- a/adapters/ts/src/index.ts
+++ b/adapters/ts/src/index.ts
@@ -222,6 +222,8 @@ export type {
   MemoryEmbedStatsResponse,
   MemoryReembedConfigured,
   MemoryReembedResponse,
+  MemoryBackfillResponse,
+  MemoryPurgeResponse,
   // Interruption
   InterruptListResponse,
   InterruptRow,
--- a/packages/memory-view/package-lock.json
+++ b/packages/memory-view/package-lock.json
@@ -1,15 +1,15 @@
 {
   "name": "@loomcycle/memory-view",
-  "version": "0.1.0",
+  "version": "0.2.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@loomcycle/memory-view",
-      "version": "0.1.0",
+      "version": "0.2.0",
       "license": "Apache-2.0",
       "devDependencies": {
-        "@loomcycle/client": "^1.49.0",
+        "@loomcycle/client": "^1.55.0",
         "@types/node": "^22.0.0",
         "@types/react": "^19.0.0",
         "@types/react-dom": "^19.0.0",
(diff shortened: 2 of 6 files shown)


denn-gubsky:
Pushed one more commit: **the dry-run test was one-directional and my own fail-before probe caught it.**

It checked backfill with `dryRun: true` and purge with `dryRun: false`, so a layer that hardcoded `true` on the backfill path would have passed. Both directions are now asserted for both ops — the two failures aren't equally bad: a commit that silently becomes a plan does nothing, while a **plan that silently becomes a commit changes data with no preview**.

Verified fail-before: forcing `dryRun: true` now fails for each op.

PR #1064, opened 2026-08-27
feat(bench): add the LoCoMo memory-retrieval benchmark harness

denn-gubsky:
## Why

loomcycle's memory has had no retrieval-quality number. `memory-eval` reports precision/recall, but only against a hand-written 20-row seed corpus embedded by a deterministic bag-of-tokens stub — so it validates the metric math and nothing about retrieval. Nobody could answer "how good is recall, and did that change make it worse".

LoCoMo closes that gap cheaply because of one property: its QA annotations carry an `evidence` field naming the `dia_id`s of the turns that support each answer. Keying every ingested turn by its `dia_id` turns each annotation into **retrieval ground truth**, so the retrieval axis needs no LLM judge and no answer-string matching — deterministic, re-runnable, and it measures the ranker and the embedder rather than the answering model.

That is also the axis worth measuring. LoCoMo's *answer* accuracy is near-saturated (a plain full-context baseline beats reported memory systems, ~73 vs ~68 J); its retrieval is not. And the corpus is small — ~5,900 turns, so a full run costs ~5,900 embeddings and nothing else.

## What

`bench/cmd/locomo` drives an externally-running instance, the same posture as `lc-bench`, so numbers come from the real vector store and the real configured embedder. Modes: `convert` (emit one `memory-eval` dataset per conversation, no instance needed), `ingest`, `search`, `all`, `purge`.

Three decisions the numbers depend on, each with a test:

- **One memory `scope_id` per conversation.** `dia_id`s are numbered per conversation and collide heavily across them — the released file has 5,882 turns and only 1,033 distinct ids — so a shared keyspace would overwrite rows *and* let a probe retrieve another conversation's turns.
- **Session timestamps are prefixed onto each row.** LoCoMo dates live on the session, not the turn; without them the temporal category is unretrievable however good the embedder is.
- **The harness resolves `GET /v1/_me` before writing and refuses the default/legacy tenant.** The memory routes derive the tenant from the principal and honour no `?tenant=` override, so a dedicated tenant bearer is the only isolation available. `-allow-shared-tenant` is the deliberate override.

Dataset defects are counted and printed rather than silently absorbed — a benchmark that quietly shrinks its answer key reports a number against a smaller key than it claims. On the current file: 1 unreadable evidence fragment, 3 evidence ids naming no turn in their own conversation, 5 questions with no usable evidence, 16 sessions declared with a timestamp and no turns. Category 5 is excluded by default (444 of its 446 questions have a `null` answer; Mem0 and Zep excluded it too), and **every report states its category filter**, because differing filters are why published LoCoMo numbers cannot be compared with one another.

The dataset is **not vendored** — LoCoMo is CC BY-NC 4.0, so `-data` takes an operator-supplied copy and the test fixture is hand-written rather than excerpted. `bench/results/` was already gitignored.

## What was tested

`go test ./...` clean (92 packages ok, 0 failures). 25 unit tests in the new package:

- `TestParseEvidence_RecoversTheMalformationsTheReleasedFileContains` — the released file's evidence field is dirty: `"D8:6; D9:17"`, `"D9:1 D4:4 D4:6"`, `"D:11:26"`, a bare `"D"`. The parser splits packed entries, repairs the unambiguous transposed colon, and reports the rest instead of guessing.
- `TestParse_ScopeIDIsPerConversationSoCollidingDiaIDsStaySeparate` — the collision guard.
- `TestParse_EvidenceNamingNoTurnInItsConversationIsDroppedAndCounted`, `TestParse_QueryWithoutUsableEvidenceIsDroppedAndCounted`, `TestParse_SessionDeclaredWithoutTurnsIsCounted`, `TestParse_CategoryFilterExcludesAdversarialAndCountsIt`
- `TestConnect_RefusesTheDefaultTenantSoTheCorpusCannotLandInRealMemory` (+ the explicit-override and no-bearer cases)
- `TestPutEntry_EscapesTheColonInADiaIDKeyAndSendsAJSONValue` — `dia_id`s contain a colon and go into the URL path.
- `TestConvert_EmitsADatasetTheRealEvalLoaderAccepts` — asserts `memory-eval`'s **own** parser reads the emitted JSONL, so the format cannot drift.
- Metric math incl. MRR, hit-rate, top-k truncation, empty-is-zero-not-NaN.

Fail-before verified on the three load-bearing ones (naive evidence parser, shared `scope_id`, removed tenant guard) — each fails on the broken code and passes on the fix. The shared-`scope_id` break also showed the hazard concretely: the converted corpus collapsed from 3 rows to 1.

Manual, against the real `locomo10.json`: converts to 10 datasets, **5,882 turns / 1,535 scoreable queries**, and `loomcycle memory-eval --dataset locomo-conv-30.jsonl` runs one (recall@10 0.1615 on the stub embedder — the lexical floor, as expected).

## Not covered

LoCoMo does not test knowledge updates, which is what the bi-temporal fact tier exists for. That axis needs a different corpus — see `/loomcycle/docs/factual-corpus-sources-and-memory-benchmarks` in the document store.

### DIFF
--- a/bench/cmd/locomo/README.md
+++ b/bench/cmd/locomo/README.md
@@ -1,0 +1,152 @@
+# locomo — LoCoMo long-term conversational memory benchmark
+
+Scores loomcycle's memory **retrieval** against the LoCoMo benchmark, on a
+corpus held in an isolated tenant. It drives an externally-running instance, so
+the numbers come from the real vector store and the real configured embedder —
+not from a harness substitute for either.
+
+## Why retrieval, and why this benchmark
+
+LoCoMo's QA annotations carry an `evidence` field: the `dia_id`s of the
+dialogue turns that support each answer. Ingest every turn as a memory row
+**keyed by its `dia_id`**, and each annotation becomes retrieval ground truth:
+
+```
+question  ->  "When did Caroline go to the LGBTQ support group?"
+expected  ->  ["D1:3"]
+```
+
+So the retrieval axis needs **no LLM judge and no answer-string matching**.
+It is deterministic, cheap and re-runnable, and it measures the ranker and the
+embedder rather than the answering model. That matters because LoCoMo's *answer*
+axis is close to saturated (a plain full-context baseline beats reported memory
+systems) while its retrieval axis is not.
+
+The corpus is also small: ~5,900 turns across 10 conversations, so a full run
+costs ~5,900 embeddings and nothing else.
+
+## The dataset is not in this repo
+
+LoCoMo ships under **CC BY-NC 4.0**. Nothing derived from it is committed here.
+Download your own copy:
+
+```sh
+curl -sSLO https://raw.githubusercontent.com/snap-research/locomo/main/data/locomo10.json
+```
+
+Pass it with `-data`. Keep the file and any generated corpus out of version
+control.
+
+## Isolate the memory first
+
+The memory routes derive the tenant from the **principal** — they honour no
+`?tenant=` override — so the only way to keep ~5,900 synthetic conversational
+rows out of real memory is a dedicated tenant's bearer.
+
+Mint one (admin bearer required), then point the harness at it:
+
+```sh
+# 1. mint an OperatorTokenDef for a tenant of its own
+curl -sS -X POST "$LOOMCYCLE/v1/_operatortokendef" \
+  -H "Authorization: Bearer $LOOMCYCLE_AUTH_TOKEN" \
+  -H 'Content-Type: application/json' \
+  -d '{"op":"create","name":"locomo-bench","tenant_id":"locomo-bench",
+       "subject":"bench","scopes":["substrate:tenant"]}'
+
+# 2. give it to the harness (preferred over reusing the operator bearer)
+export LOOMCYCLE_LOCOMO_TOKEN='<the minted token>'
+```
+
+The harness calls `GET /v1/_me` before it writes anything, prints the tenant it
+resolved, and **refuses to run** if that tenant is the default/legacy one.
+`-allow-shared-tenant` overrides this deliberately for single-tenant
+deployments.
+
+Prerequisites on the instance: a vector-capable store
+(`LOOMCYCLE_PGVECTOR_ENABLED=1` on Postgres, or a sqlite-vec build) and a
+configured `memory.embedder`. Without either, ingest aborts on the first embed
+warning rather than writing thousands of un-embedded rows.
+
+## Modes
+
+```sh
+go build -o bin/locomo ./bench/cmd/locomo
+
+# offline: emit one memory-eval dataset per conversation (no instance needed)
+./bin/locomo -mode=convert -data locomo10.json -out /tmp/locomo-datasets
+loomcycle memory-eval --dataset /tmp/locomo-datasets/locomo-conv-26.jsonl
+
+# live: ingest, then score
+./bin/locomo -mode=all -data locomo10.json -loomcycle http://127.0.0.1:8787
+
+# smoke it on one conversation first
+./bin/locomo -mode=all -data locomo10.json -conversations 1
+
+# reclaim the scope
+./bin/locomo -mode=purge -data locomo10.json
+```
+
+`convert` output runs through `loomcycle memory-eval`, but that command embeds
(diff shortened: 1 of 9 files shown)


denn-gubsky:
## First real retrieval baseline

Ran the full benchmark end to end against a throwaway local instance — Postgres 17 + pgvector 0.8.6, embedding with **bge-m3** (1024-dim) via Ollama, under an isolated `locomo-bench` tenant minted for the run.

**5,882 rows ingested and embedded, 0 embed warnings. 1,535 queries scored. Categories 1–4 (5 excluded).**

| slice | queries | recall@10 | MRR | hit-rate | p50 ms | p99 ms |
|---|---|---|---|---|---|---|
| **overall** | **1535** | **0.6675** | **0.5029** | **0.7407** | 149.6 | 218.4 |
| single-hop | 841 | 0.7446 | 0.5175 | 0.7586 | 150.2 | 217.6 |
| temporal | 320 | 0.7500 | 0.5988 | 0.7781 | 149.0 | 202.2 |
| multi-hop | 282 | 0.4359 | 0.4220 | 0.7199 | 148.1 | 228.1 |
| open-domain | 92 | 0.3856 | 0.2843 | 0.5109 | 148.0 | 252.8 |

Three things this says, beyond "it runs":

- **The session-timestamp decision is load-bearing and now has evidence.** Temporal scores 0.750 recall / 0.779 hit-rate — level with single-hop. Without the session date prefixed onto each row it would be near zero, because LoCoMo puts the date on the session and not the turn.
- **Multi-hop's gap is recall, not retrieval.** hit-rate 0.720 against recall 0.436 means the store reliably finds *one* of the several turns that support a multi-hop answer and misses the rest. That is the actionable signal, and it is only visible because hit-rate is reported next to recall.
- **Open-domain is the floor at 0.386 / 0.511**, as expected — those questions draw on world knowledge rather than a stored turn, so there is a ceiling no embedder clears.

Latency is flat at ~150 ms p50 across every slice, which is the embed round trip to a remote Ollama dominating, not the vector search.

Caveat on comparability: this is a **retrieval** number against the evidence answer key. It is not the answer-accuracy figure vendors publish for LoCoMo and should not be set beside those.

The `matrix.md` above was rendered by the binary built before 5a37161, so its defect heading lacks the population label that commit adds; the numbers are unaffected.

denn-gubsky:
## Reproduced on the TrueNAS instance — identical quality, lower latency

Second run, this time against the real deployment (`192.168.0.77:8787`) under an operator-minted `locomo-bench` tenant token (`substrate:tenant`, non-admin), with the same dataset. That instance runs the same embedder — `ollama-local` / `bge-m3:latest`, 1024-dim — so the two runs are directly comparable.

**5,882 rows ingested and embedded, 0 embed warnings. 1,535 queries scored.**

| slice | queries | recall@10 | precision@10 | MRR | hit-rate | p50 ms | p99 ms |
|---|---|---|---|---|---|---|---|
| overall | 1535 | 0.6675 | 0.0857 | 0.5029 | 0.7407 | **96.9** | **104.9** |
| single-hop | 841 | 0.7446 | 0.0779 | 0.5175 | 0.7586 | 97.2 | 104.6 |
| temporal | 320 | 0.7500 | 0.0828 | 0.5988 | 0.7781 | 96.3 | 104.6 |
| multi-hop | 282 | 0.4359 | 0.1209 | 0.4220 | 0.7199 | 96.4 | 103.7 |
| open-domain | 92 | 0.3856 | 0.0587 | 0.2843 | 0.5109 | 97.1 | 109.1 |

**Every quality metric matches the local run to four decimal places, on every slice.** Two entirely different deployments — a throwaway local Postgres 17 + pgvector 0.8.6 versus the TrueNAS instance's own store — produced bit-identical recall, precision, MRR and hit-rate. That is the reproducibility claim the harness needs in order to be a gate: a number that moves means the change moved it, not the environment.

Latency is the only thing that differs, and it improved: p50 150 → 97 ms, p99 218 → 105 ms, with much tighter spread. Expected — the embed round trip is now local to the box running Ollama instead of crossing the LAN from a laptop, which also removes the variance the p99 was measuring.

Note the corpus is left in place in the `locomo-bench` tenant so `-mode=search` can be re-run without paying for 5,882 embeddings again; `-mode=purge` reclaims it.

PR #1151, opened 2026-09-09
ci: gate the coordination tier on the Postgres job

denn-gubsky:
Closes the CI blind spot named in #1150.

## Why

Every test in `./internal/coord/...` is PG-gated, and **nothing ran them**:

- the `Go` job invokes `go test ./...` with no `LOOMCYCLE_TEST_PG_DSN`, so they all skip
- this job's store step covers only `./internal/store/...`

So the cancel/steer coordinators, the Postgres backplane, the replica store + heartbeat, the replicas sweeper and user quotas were structurally invisible to CI. And it drifted, as invisible things do: six tests sat broken on `main` (fixed in #1150) because their fixtures built replica ids containing a `.`, which `ValidateReplicaID` rejects. Nothing could go red.

## What

Two steps on the existing `go-postgres` job.

**A separate database — not the store contract's, and not a convenient ordering.** This is the part worth reviewing. The coord tests need a **migrated** `public` schema (the store contract creates a per-test schema and deliberately leaves `public` empty). But two store tests need `public` **left un-migrated**:

- `TestOpen_ExtensionPresentButTableMissingDisablesVectorSupport`
- `TestMigrate0062_RepairedTableMatchesFromScratchShape` — fails `0062 did not create memory_embeddings`

The two requirements are mutually exclusive in one database. Sharing `loomcycle_test` would work only while the steps happen to run in this order, and would break on a reorder or a re-run — with a failure that reads like a migration bug rather than a CI-config problem. Same reasoning as the `sqlmem_aux` DB already in this job.

**Migrated with loomcycle's own migrator** (`migrate up --dsn`) rather than a psql loop over `migrations/*.up.sql`, so the ordering cannot drift from the real one and `schema_migrations` is stamped as in production.

**`-race`**, because this package is entirely about concurrency — a backplane fanning out to subscribers, a heartbeat goroutine, a sweeper racing live replicas. That's the bug class that matters here, and the suite is fast enough (~3 s) to afford the detector.

**Job renamed to "Go (Postgres tiers)"** — it gates three PG-only surfaces now, and a red check named "store contract" would send the next person debugging a coord or sqlmem regression to the wrong package. The job key stays `go-postgres`; `main` has no branch protection, so no required-check name breaks.

## What was tested

A CI-config change is only as good as the evidence it actually runs, so I ran both steps **verbatim** against local PostgreSQL 17 — as the `loomcycle` role, exactly as CI does, not as a superuser:

```
### Provision the coordination-tier DB ###
CREATE DATABASE
CREATE EXTENSION
migrate up: OK (version=74 dirty=false)
### Coordination tier (Postgres) ###
ok  github.com/denn-gubsky/loomcycle/internal/coord  3.134s
```

**The gate is not decoration.** Reintroducing the `.` that #1150 removed turns the new step red:

```
--- FAIL: TestCancelCoordinator_NotFound_ReturnsMiss
--- FAIL: TestCancelCoordinator_DeadOwner_MarksRunFailed
--- FAIL: TestCancelCoordinator_Timeout_ReturnsOwnerUnreachable
--- FAIL: TestPostgresBackplane_PublishSubscribeRoundtrip
--- FAIL: TestPostgresBackplane_SelfMessageFiltered
```

**No interference.** The store contract's two repair tests still pass against their own un-migrated database, confirming the separate DB leaves them alone.

YAML parses (`yaml.safe_load`) and the step order is valid — `psql` is installed by an earlier step in the same job.

## Note on runtime

This adds ~3 s of tests plus a `go run` of the migrator and two `psql` calls to a job that already runs ~10 min. `-race` on this package is cheap; I did not add it to the store contract, which is the slow step.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -187,7 +187,15 @@ jobs:
       # Migrated with loomcycle's OWN migrator rather than a psql loop over
       # migrations/*.up.sql, so the ordering cannot drift from the real one and
       # schema_migrations is stamped as in production.
+      #
+      # !cancelled() so a red store contract does not SKIP this gate. That is
+      # not hypothetical: on this step's first CI run the store contract
+      # exhausted its 1800s budget and both coord steps were skipped, which
+      # would have silently restored the very blind spot they exist to close.
+      # The two surfaces share only the service container, so coord's result is
+      # independent and worth having even when the contract is red.
       - name: Provision the coordination-tier DB
+        if: ${{ !cancelled() }}
         env:
           PGPASSWORD: loomcycle
         run: |
@@ -203,6 +211,7 @@ jobs:
       # live replicas); a data race here is the bug class that matters, and the
       # suite is fast enough (~3s locally) to afford the detector.
       - name: Coordination tier (Postgres)
+        if: ${{ !cancelled() }}
         env:
           LOOMCYCLE_TEST_PG_DSN: postgres://loomcycle:loomcycle@127.0.0.1:5432/coord_test?sslmode=disable
         run: go test -count=1 -race -timeout 600s ./internal/coord/...
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,7 +82,12 @@ jobs:
           fi
 
   go-postgres:
-    name: Go (Postgres store contract)
+    # Named for the TIERS it gates, not just the store contract: this job now
+    # owns three PG-only surfaces (store contract, coordination tier, SQL
+    # Memory), and a red check named "store contract" sends the next person
+    # debugging a coord or sqlmem regression to the wrong package. The job KEY
+    # stays go-postgres.
+    name: Go (Postgres tiers)
     runs-on: ubuntu-latest
     # Runs the shared storetest contract against a REAL Postgres so
     # backend-parity bugs are caught in CI, not only on SQLite. The contract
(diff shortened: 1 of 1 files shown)


denn-gubsky:
CI evidence from the first run, and one fix it prompted.

## The gate works — 28 previously-skipped tests now run

```
Provision the coordination-tier DB   success
  CREATE DATABASE / CREATE EXTENSION / migrate up
Coordination tier (Postgres)         success
  ok  github.com/denn-gubsky/loomcycle/internal/coord  2.930s
```

`2.930s` is the load-bearing number, because a step that skips everything also goes green. Locally, **without** the DSN — which is what CI did until now — the package reports `ok ... 0.786s` with **28 `--- SKIP` lines**. With the DSN it takes 2.930s in CI, matching the 3.134s I measured locally. So those 28 tests are genuinely executing, not passing vacuously. For contrast, the same CI log shows what a non-run looks like: `?  internal/store/storetest  [no test files]`.

## The first run failed, and it was not this change

The pre-existing "Postgres store contract" step hit `panic: test timed out after 30m0s`. Not a hang — `TestStoreContractWithPgvector` was 13m4s in with a normal 3s subtest running; the 1800s budget was simply exhausted, which is the confusing failure mode that step’s own comment warns about.

Variance, confirmed by measurement rather than assumed:

| run | branch | pg job |
|---|---|---|
| 34311183467 (first) | this branch | **31m30s FAIL** |
| 34311183467 (re-run, same commit) | this branch | **9m4s PASS** |
| 34310966746 | main | 8m41s |
| 34310944459 | main | 9m31s |
| 34262672069 | older | 18m6s |

Main runs the job in ~9m *with the new store-contract tests already merged*, so those are not the cause; this branch differed from main only by `ci.yml`, which cannot slow a step it does not touch; and the 18m historical outlier shows the job already has wide variance.

## What the failure did expose — fixed in 097ad762

Both coord steps were **skipped**, because an unrelated step failed first. That is a genuine flaw in how I wired this, independent of the timeout: a gate meant to make coord regressions visible must not vanish exactly when the job is least healthy — it would silently restore the blind spot it exists to close. Added `if: ${{ !cancelled() }}` to both steps (not `always()`; a cancelled run should stop, not go on to provision a database). The two surfaces share only the service container, so coord’s verdict is independent.

## Two things flagged, not folded in

- **The sqlmem steps have the identical coupling** — also skipped in that run. Same one-line guard fixes them; left as a separate concern.
- **The 1800s budget has thin headroom**: ~9m baseline, 18m observed, 30m+ once. Worth raising it, or splitting the two contract variants into separate steps, so a slow runner produces a clear failure instead of a misleading one.

PR #1179, opened 2026-09-10
feat(memory): read facts from the chunk plane, by class rather than namespace

denn-gubsky:
RFC CV **P2d** — the read move.

> ⚠️ **Stacked on #1177** (P2b). Without the origin stamp on a fact's body row, provenance selects nothing in the chunk plane and this change is merely inert. **Merge #1177 first**; this diff narrows to its own two files once that lands.

## Why

Once a fact's home is a chunk, *"everything except prose"* stops being expressible as *"not the chunk namespace"* — because the facts now live **inside** that namespace.

- `sources=facts` excluded the namespace, and would have hidden exactly what was asked for.
- `sources=facts,notes` — the **recall default** — excluded it too, and would have hidden every fact in the store.

## What

The selectors are now defined by **class**, matching the classifier they have to agree with:

| source | predicate |
|---|---|
| `facts` | provenance present, in **either** namespace |
| `notes` | no provenance **and** not a chunk body |
| `documents` | a chunk body **and** no provenance |
| `facts+notes` | everything except the document **class** |

The last needs a predicate the filter did not have. **`ExcludeDocumentPrefix`** drops rows that are chunk bodies *without* provenance — exactly the document class and nothing else.

It is deliberately **not** a variant of `ExcludeKeyPrefix`: that field drops a *namespace*, this one drops a *class*, and conflating them is what made the old encoding wrong. It stays a prefix rather than a bool because the store owns no knowledge of what the document namespace is called — the caller supplies it, as it already does for the other two prefix fields.

Applied to **both** search legs: the vector and lexical legs are fused by RRF, so a row admitted by either reaches the caller, and filtering one filters nothing.

## A latent bug found on the way

`MemorySearchFilter.IsZero` documents itself as *"the filter constrains nothing — the signal that a backend may take a query path with no predicate at all"* while testing only the two prefix fields. A provenance-constrained filter therefore reported "constrains nothing", and a backend that trusted it would have returned notes alongside facts. Latent because no backend consumes it yet — but the contract invites exactly that path, and adding a third predicate field made it worse. Every predicate-bearing field is now listed.

## What was tested

Against **real Postgres with pgvector**, not skipped — the whole point, since these paths are postgres-only and a sqlite run would have passed vacuously. The store contract gains an `ExcludeDocumentPrefix` case whose distinguishing row is a chunk body **with** provenance: excluding the namespace drops it, the class predicate must keep it. Asserted on both legs.

Verified **non-vacuous** by degrading the predicate to a namespace exclusion, which fails with *"dropped a chunk body that HAS provenance … or a collapsed fact becomes unreachable by the recall default"*.

Full `go test ./...` clean; full pgvector contract suite green (126s).

## Not in this PR

P2d's remaining half — the operator listing going chunk-backed, the Web UI moving with it, and the backfill (including copying `access_count` onto the body row, per #1178) — is a separate change touching the SPA. Splitting it keeps this one reviewable as a store/filter change. The irreversible step (deleting the k/v fact row) is P2f and is not here.

### DIFF
--- a/internal/memory/backend.go
+++ b/internal/memory/backend.go
@@ -216,13 +216,25 @@ func (q SearchQuery) Filter() (store.MemorySearchFilter, error) {
 		// fact's body row now carries an origin, so excluding provenance is what
 		// makes this selector mean prose.
 		f.Provenance = store.ProvenanceAbsent
-	case memory && !docs:
+	case facts && !notes:
+		// FACTS ARE WHEREVER PROVENANCE IS, with no namespace constraint. This used
+		// to exclude the chunk namespace, which was right only while a fact's home
+		// was its own k/v row and the chunk body was a duplicate of it. A fact homed
+		// in a chunk lives at that prefix, so the exclusion would now hide exactly
+		// what was asked for.
+		f.Provenance = store.ProvenanceRequired
+	case notes && !facts:
+		// A note is what an agent wrote down ITSELF: no provenance, and not a chunk
+		// body. Both halves are load-bearing — prose has no provenance either, so
+		// provenance alone would return document text as notes.
 		f.ExcludeKeyPrefix = DocumentChunkKeyPrefix
-		if facts && !notes {
-			f.Provenance = store.ProvenanceRequired
-		} else if notes && !facts {
-			f.Provenance = store.ProvenanceAbsent
-		}
+		f.Provenance = store.ProvenanceAbsent
+	case memory && !docs:
+		// facts + notes — everything EXCEPT prose, and the default for recall.
+		// No longer expressible by excluding the chunk namespace, because the facts
+		// half now lives inside it; this excludes the document CLASS, which is the
+		// only thing being ruled out.
+		f.ExcludeDocumentPrefix = DocumentChunkKeyPrefix
 	}
 	// docs && facts && notes → everything; the zero filter already says that.
 	return f, nil
--- a/internal/memory/sources_test.go
+++ b/internal/memory/sources_test.go
@@ -45,15 +45,14 @@ func TestSearchQueryFilter_SourcesMapToPredicate(t *testing.T) {
 				"(exclude AND require the same prefix), which would return nothing",
 		},
 		{
-			name: "facts only requires provenance",
+			name: "facts are wherever provenance is, in EITHER namespace",
 			q:    SearchQuery{Sources: []Source{SourceFacts}},
-			want: store.MemorySearchFilter{
-				ExcludeKeyPrefix: DocumentChunkKeyPrefix,
-				Provenance:       store.ProvenanceRequired,
-			},
+			want: store.MemorySearchFilter{Provenance: store.ProvenanceRequired},
 			comment: "origin is server-stamped, so it is the unforgeable discriminator " +
 				"(RFC BW §9 Q1) — class is model-supplied and would let an agent promote " +
-				"its own note to a fact",
+				"its own note to a fact. No namespace constraint: excluding the chunk " +
+				"prefix was right only while a fact's home was its own k/v row and the " +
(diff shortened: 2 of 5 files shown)


denn-gubsky:
## Review-pass findings on my own change

Two things came out of the self-review. One is fixed here; the other changes the **merge order** and is the more important of the two.

### Fixed: a body rewrite was erasing a fact's provenance

`acdfab56`. `update_chunk` rewrites a chunk body through the origin-less path, and a plain `MemorySet` delegates to `MemorySetProvenance` with the extra arguments zeroed — so it **clears** the column. Verified against the store, not assumed: origin after a plain set over a provenanced row reads `""`.

Once that column is what tells a fact from prose, an operator merely correcting a fact's wording would silently reclassify it as a document — the text still right, the fact just no longer a fact, gone from `sources=facts` and answering as prose. `writeBodyAs` now preserves what it is not told, the same rule `writeChunkMeta` already applies to the sidecar's source span. Regression test fails without it with `origin "operator" -> ""`.

### ⚠️ Merge order: this should land WITH P2f, not before it

`sources=facts` is now `provenance present, no namespace constraint`. During the dual-write window a fact exists **twice** — its `memory/<class>/<slug>` row and its `doc.chunk:<hex>` body — and after P2b both carry an origin. So both match, and **recall returns every fact twice**.

Reasoned from the filter rather than measured, and the store bears out the shape: of 84 facts in a live scope, 84 `memory/` rows carry an origin, and their bodies read 0 only because they predate the stamp — with it they match too.

Before this change `sources=facts` excluded the chunk namespace, so there was no duplicate. This change removes that exclusion, which is correct for the end state and wrong for the window. Duplicates halve the effective top-k, which is an accuracy effect in a phase whose gate explicitly forbids one.

So: **merge #1177 → then this together with P2f (the k/v delete)**, or hold this until P2f is ready. Landing it alone is the one ordering that regresses recall. I'd rather say so than let it merge on the strength of a green suite — the suite does not cover the two-plane duplication, because the fake toolset writes one plane.

An exhaustive check of all 8 source combinations is otherwise clean: `[facts]` provenance-only, `[notes]` excludes namespace + provenance-absent, `[facts,notes]` excludes the document class, `[documents]` namespace + provenance-absent, the two mixed-with-documents sets still correctly refused, and `[]`/all-three both zero.

denn-gubsky:
## Rebased — conflicts resolved, and the review fix split out

The conflict was the stacked-PR one: this branch still carried #1177's original commit, which `main` now has as a squash. Rebased with `--onto`, dropping that duplicate. **No content conflicts** — it was purely the duplicated commit. This PR is now a clean five-file diff of its own change, and GitHub reports it mergeable.

**The review fix moved to #1188.** `fix(memory): a body rewrite must not erase a fact's provenance` was committed here, but it repairs a bug that #1177 put **on `main`** — `update_chunk` strips a fact's provenance and silently reclassifies it as prose. Leaving it in this PR would have held the fix behind the merge-ordering constraint below, which is the wrong trade for a live bug. Its fail-before now runs against current `main`'s own `document.go` rather than a synthetic revert, which is what shows it is live rather than hypothetical.

### The merge-ordering constraint still stands

Repeating it because it survived the rebase and is the thing most likely to be missed: **this should land with P2f, not before it.**

`sources=facts` becomes "provenance present, any namespace". During the dual-write window a fact exists twice — its `memory/<class>/<slug>` row and its `doc.chunk:` body — and after #1177 both carry an origin. So both match and **recall returns every fact twice**, halving the effective top-k. That is an accuracy effect in a phase whose gate forbids one.

Green CI is not evidence against this: the bundle's fake toolset writes one plane, so the suite cannot see the duplication.

Re-verified after the rebase against real Postgres with pgvector (main has moved several PRs since this was written), plus `go test ./...` clean.

PR #1180, opened 2026-09-10
feat(channels): a held channel stores without delivering, until released

denn-gubsky:
**RFC CY item 9 (phase L5)** — the breakpoint a workflow needs.

## Why

A workflow wired through channels has **no pause**. Once a stage publishes, the next stage's subscriber wakes and the wave is gone — there is nowhere to stand between "the results are in" and "the next wave starts". An operator who wants to look before the run continues has only bad options today: stop the whole runtime, or let it go and read the transcript afterwards.

`hold: true` on a channel definition makes it a **breakpoint**. A publish is stored and reported, but delivered to nobody and waking nobody, until someone releases it — `Channel op=release` for an agent, `POST /v1/_channels/{name}/release` for an operator, a **Release** control on the channel page for a human. `count` defaults to 1, oldest first, so a workflow can be **single-stepped**: the wave upstream runs, its results queue, and nothing downstream starts until someone says go.

```
POST /v1/_channels/review-queue/release   { "count": 1 }
  → { "channel": "review-queue", "released": ["msg_…"], "released_count": 1, "still_held": 2 }
```

## A reserved `visible_at`, not a `held` column

Every channel read already filters `visible_at <= now`. Marking a held message with a reserved far-future instant (`store.ChannelHeldVisibleAt`, 2200-01-01 — inside int64 nanos, which run out in 2262) withholds it through machinery that **already exists**: subscribe, peek, await, and any read path added later, without each one having to remember.

A boolean column would need its own predicate on every one of those queries, and the query that *forgets* it delivers the message the hold was supposed to stop. **The two designs fail in opposite directions and only one of them is safe.** It also costs no schema change on `channel_messages`, and the existing overflow trim already orders by `(visible_at, id)`.

## The check lives in the publisher

`StorePublisher.HoldFn` — wired once in `main.go`, resolving through a point lookup — means every internal publish path (heartbeats, the webhook relay, interrupts, the admin endpoint) honours a hold without each call site remembering to ask. *A promise enforced at five call sites is a promise the sixth one breaks.* The in-band tool checks the def it already resolved; the connector passes the reserved instant so the publisher short-circuits rather than resolving twice.

## Release is gated by publish, not subscribe

Releasing is the act of making a message deliverable — the half of a publish the hold deferred — so the question is whether the agent may *put messages on* this channel, not whether it may read them. The `_system/` refusals apply for the same reason.

## What a hold does not change

- **TTL still counts from publish time.** An expired held message is never released and never delivered — holding is not a way to outlive the retention its publisher declared.
- **Overflow still trims the oldest**, reporting `dropped_oldest`. A hold buffers; it does not make the buffer unbounded.
- **The hold wins over `deliver_at`** — a held message waits for a release, not for a clock — so the publish result says `"held": true` rather than promising a `visible_at` in the year 2200.

Releasing a channel with nothing held reports **zero rather than failing**, and a channel switched back to `hold: false` can still release what it holds — turning the breakpoint off must not strand the queue with no way out but a purge.

## The definition plane, end to end

`hold` is carried by operator yaml, the `channels` table (sqlite ALTER + postgres **migration 0075**, default false so every existing channel delivers exactly as before), `ChannelPatch`, the substrate CRUD, the MCP `channeldef` schema, `Context op=channels`, the TS adapter (`createChannel`/`updateChannel`/`releaseChannel`), and the Web UI (a hold checkbox, a hold chip, a Release control that appears only on a held channel). *A def field that survives only half the plane is a setting the operator believes they made.*

## Tested

`go test ./...` clean (**100 packages**, full output grepped for `FAIL`) plus the **postgres tier** against a local Postgres 14 (`TestStoreContract`, 128 s — migration 0075 and the `FOR UPDATE` release path both exercised on the real backend).

- **store contract, both backends** — hold round-trips through create/list/patch and an unrelated patch does not clear it; release hands over oldest first, reports what is left, drains, and is a no-op on an empty queue; an expired held message is neither released nor counted;
- **the tool** — a held publish is stored, reports `held:true`, and is *not* delivered, with a control channel in the same fixture proving the fixture otherwise delivers; release is refused with a subscribe-only ACL and the message stays held; TTL expiry survives the hold;
- **the wire** — the admin publish route holds, the release endpoint steps oldest first, an undeclared channel 404s, a `deliver_at` cannot escape the hold, and a **runtime-declared (substrate)** channel holds and releases — the path the canvas will use, which the yaml map alone would not cover;
- **the seam** — a publisher that never resolves a definition (`SystemPublisher.PublishNow`, the heartbeat/webhook/interrupt shape) is held anyway.

**The fail-before pass found a hole in its own tests** and the second commit closes it: with the `HoldFn` consultation removed, every HTTP hold test still passed, because they only exercised the connector's own def-aware check. The seam test now fails on exactly that deletion (`an internal publish walked past the hold: 1 delivered, want 0`). Reverting the tool's hold arm fails three tool tests; reversing the release ordering fails the store contract.

## Known gaps, deliberate

- The **webhook channel relay** does not resolve a channel def at all today (it passes `0/0` for `max_messages` and `default_ttl` too), so a hold does not apply there yet — it arrives with the L4 trigger work.
- **gRPC's `ListChannels`** descriptor does not carry `hold`: gRPC has no channel CRUD, and regenerating the pb for one read-only field is not worth it in this change.

### DIFF
--- a/adapters/ts/src/client.ts
+++ b/adapters/ts/src/client.ts
@@ -79,11 +79,13 @@ import type {
   ChannelPeekResult,
   ChannelPublishResult,
   ChannelPurgeResult,
+  ChannelReleaseResult,
   ChannelSubscribeResult,
   CreateChannelOptions,
   ListChannelsResponse,
   PeekChannelOptions,
   PublishChannelOptions,
+  ReleaseChannelOptions,
   SetMemoryEntryOptions,
   SetMemoryEntryResponse,
   SubscribeChannelOptions,
@@ -2390,6 +2392,7 @@ export class LoomcycleClient {
     if (opts.max_messages !== undefined) body.max_messages = opts.max_messages;
     if (opts.publisher !== undefined) body.publisher = opts.publisher;
     if (opts.period !== undefined) body.period = opts.period;
+    if (opts.hold !== undefined) body.hold = opts.hold;
     return postJSON<ChannelDescriptor>(this.ctx, "/v1/_channels", body, {
       signal: opts.signal,
     });
@@ -2407,6 +2410,7 @@ export class LoomcycleClient {
     if (opts.default_ttl !== undefined) body.default_ttl = opts.default_ttl;
     if (opts.max_messages !== undefined) body.max_messages = opts.max_messages;
     if (opts.semantic !== undefined) body.semantic = opts.semantic;
+    if (opts.hold !== undefined) body.hold = opts.hold;
     return patchJSON<ChannelDescriptor>(
       this.ctx,
       `/v1/_channels/${encodeURIComponent(name)}`,
@@ -2444,6 +2448,27 @@ export class LoomcycleClient {
       this.ctx,
       `/v1/_channels/${encodeURIComponent(name)}/purge`,
       {},
+      { signal: opts?.signal },
+    );
+  }
+
+  /** Hand the oldest `count` (default 1) messages held on a `hold:`
+   *  channel to its subscribers. Allowed on yaml-declared channels —
+   *  releasing moves messages, it does not mutate the definition.
+   *  Releasing a channel with nothing held reports zero rather than
+   *  failing. */
+  async releaseChannel(
+    name: string,
+    opts?: ReleaseChannelOptions,
+  ): Promise<ChannelReleaseResult> {
+    const body: Record<string, unknown> = {};
+    if (opts?.count !== undefined) body.count = opts.count;
+    if (opts?.scope !== undefined) body.scope = opts.scope;
+    if (opts?.scope_id !== undefined) body.scope_id = opts.scope_id;
+    return postJSON<ChannelReleaseResult>(
+      this.ctx,
+      `/v1/_channels/${encodeURIComponent(name)}/release`,
+      body,
       { signal: opts?.signal },
     );
   }
--- a/adapters/ts/src/index.ts
+++ b/adapters/ts/src/index.ts
@@ -286,6 +286,7 @@ export type {
   ChannelPeekResult,
   ChannelPublishResult,
   ChannelPurgeResult,
+  ChannelReleaseResult,
   ChannelScope,
   ChannelSubscribeResult,
   PeekChannelOptions,
@@ -301,6 +302,7 @@ export type {
   ChannelBroadcastResult,
   // Channel admin CRUD (v0.11.5)
   CreateChannelOptions,
+  ReleaseChannelOptions,
   UpdateChannelOptions,
   // Memory entry admin CRUD (v0.11.5)
   SetMemoryEntryOptions,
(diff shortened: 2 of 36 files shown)


denn-gubsky:
**Self-review pass** — three findings, all fixed in `fix(channels): keep the reserved held instant out of operator-facing output`.

1. **The listing printed the sentinel as a delivery time.** `GET /v1/_channels` rendered a held message's reserved `2200-01-01` `visible_at` as `newest_visible_at`, which reads as a bug rather than as "this channel is holding". Now `""` — a held message has no scheduled arrival, and the `hold` flag is what says it is held.
2. **A comment cited a drift test that did not exist** between the wire cap and the tool cap. Now it does. Also added the missing over-cap test (refused, not silently clamped).
3. **I described `ChannelHeld`'s tenant resolution wrongly** in the feature commit. It resolves yaml first (operator-global, every caller sees it) and only then the runtime row *in the caller's tenant*. So the inbound **webhook relay does honour a yaml-declared hold** — it publishes through `StorePublisher`, where the check lives — and only misses another tenant's *runtime-declared* one. The "known gap" in the PR description above is therefore too broad: what the webhook relay actually lacks is def resolution for `max_messages`/`default_ttl` and visibility of a tenant-scoped runtime channel. A test now pins all three cases instead of leaving it to prose.

`go test ./...` still clean (100 packages); fail-before verified on the listing fix.

denn-gubsky:
**Verified live** on a throwaway instance (`:8911`, its own data dir — `:8799` untouched), which is also how the last commit's defect was found.

Exercised end to end, including main.go's `HoldFn` wiring — the one line no test reaches:

- publish to the yaml-declared `hold: true` channel → `{"held":true}`, **no `visible_at`**; `peek` returns `[]` while the control channel in the same config delivers normally;
- `POST …/release` (empty body) → hands over the **oldest**, reports `still_held:1`; a second release with `{"count":10}` drains it and both messages then peek back **in publish order**;
- `GET /v1/_channels` shows `hold:true`, and with messages still held it **suppresses the reserved instant** rather than printing `2200-01-01` as a delivery time (`message_count:3, oldest_visible_at:…` and no `newest_visible_at`);
- a **runtime-declared** channel (`POST /v1/_channels` with `hold:true` — the path the canvas will use) holds and releases identically.

**What the live run found:** an over-cap `count` reached `writeChannelError`'s default arm and answered **500**, so a caller's own typo read as "the server broke". Fixed to 400 (negative too; `0` still means one), re-verified on the wire: `over-cap 400 · negative 400 · zero 200 · undeclared 404`.

denn-gubsky:
The last commit on this branch (`d150edb1`, the 400-vs-500 fix found by running the endpoint) was pushed while the squash was landing, so it is **not** on `main` — the squash took the head as it stood at three commits. Carried over as **#1185**; code unchanged.

PR #1181, opened 2026-09-10
feat(scheduler): a schedule can deliver to a channel instead of a run

denn-gubsky:
**RFC CY item 11 (phase L6)** — the time-side twin of the `delivery: channel` a webhook already has.

## Why

A workflow driven off channels needs a clock, and today reaching a channel on a cron means **burning an agent run whose only job is to publish** — a model call, a run row, and a provider bill for a message the scheduler can write itself. Webhooks solved the same problem already: `delivery: channel` relays an inbound event straight onto a channel with no run. This is that, with **time** as the event.

```yaml
scheduled_runs:
  nightly-wave:
    delivery: channel
    channel: wave-in
    schedule: "0 3 * * *"
    metadata: { batch: nightly }
```

The tick carries `{schedule_name, fired_at, delivery, payload}` and lands at the channel's **declared** scope — the same resolution an `on_complete: channel.publish` hook uses, because a global channel's tick has to be visible to a global reader, not buried under the schedule's user scope. The payload is the def's `metadata:`, the operator-authored non-secret structured data already on the def and the only thing there is to say when no agent ran, so this needs no new field.

## A tick is a fire

`next_run_at` advances, `max_fires` counts, a publish failure is recorded as `failed` with its error. **A failure counts toward the cap on purpose:** the tick happened, and a channel that is undeclared or unreachable will fail identically next time, so exempting it would let a broken schedule run forever.

## The run-shaped fields are refused, not ignored

`agent`, `prompt`, `on_complete` and credentials are **rejected** on a channel tick — at config-load *and* at substrate write. A prompt nothing sends, or an `on_complete` that can never fire because no run completes, is a setting the author believes they made; this codebase has paid for that class of bug before. Running the same rule on both write paths means an agent authoring a def cannot create a shape the operator's yaml would refuse.

`unmarshalDef`'s agent requirement becomes **per-delivery** rather than universal: a channel def legitimately has no agent, while a run def with no agent still fails at decode, where the sweeper records it as a def problem instead of discovering it inside the runner.

Carried across all three JSON mirrors — `mergedScheduleDef` (write), `lookup.SubstrateScheduleDef` (read), `scheduler.scheduleDef` (fire) — plus `config.ScheduledRun`, `ToConfigDef`, `staticToMergedScheduleDef`, the fork overlay, and the `scheduledef` tool schema. **The lookup drift guard caught the missing `want` entries**, which is what it is for.

## Two commits

1. **`refactor(scheduler)`** — pure extraction, no behaviour change: `fireOne`'s tail (advance `next_run_at`, record the result, apply `max_fires`) becomes `recordFireOutcome`. A fire that skips either half re-presents on the next tick or never retires, so the new fire kind must not reimplement it.
2. **`feat(scheduler)`** — the delivery itself.

## Tested

`go test ./...` clean (**100 packages**, full output grepped for `FAIL`).

- **scheduler** — a due channel def publishes a tick and `RunOnce` is called **zero** times (the count is the assertion: the old way to reach a channel on a cron was to burn a run); the tick's shape, scope and metadata payload are checked; an undeclared channel is a recorded failure that still advances; `max_fires: 1` retires after one tick; a disabled channel schedule publishes nothing;
- **config** — a channel tick loads with no agent, and each run-shaped field is refused by name; a missing channel is refused; `delivery: run` still requires an agent and now refuses a stray channel; an unknown delivery is refused;
- **substrate** — a channel def round-trips through create → definition JSON, and the tool refuses the same seven mixes the yaml validator does.

**Verified fail-before:** disabling the fire branch fails the scheduler tests with `RunOnce calls = 1, want 0`; deleting one validator arm fails exactly its subtest.

### DIFF
--- a/cmd/loomcycle/main.go
+++ b/cmd/loomcycle/main.go
@@ -2930,7 +2930,20 @@ func main() {
 		// Resolve on_complete: channel.publish to the channel's DECLARED scope
 		// (F37 / RFC T) so a hook on a scope:global channel lands at global,
 		// not under the run's user scope. Must be set before Start.
-		sched.SetChannelScope(srv.ResolveChannelScope)
+		// Adapt the server's channel def to the scheduler's DeclaredChannel:
+		// the two packages meet here rather than importing each other for the
+		// sake of one struct.
+		sched.SetChannelScope(func(ctx context.Context, name string) (scheduler.DeclaredChannel, bool) {
+			def, ok := srv.ResolveChannelScope(ctx, name)
+			if !ok {
+				return scheduler.DeclaredChannel{}, false
+			}
+			return scheduler.DeclaredChannel{
+				Scope:       def.Scope,
+				DefaultTTL:  def.DefaultTTL,
+				MaxMessages: def.MaxMessages,
+			}, true
+		})
 		// RFC BL P2 consolidation fan-out: the provider resolver decides
 		// parallel-vs-serial dispatch (a local model runtime is serialized), and
 		// the advisory lock makes exactly one replica per tick enumerate the
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -581,9 +581,11 @@ Operators declare run templates under the yaml `scheduled_runs:` map (`config.Sc
 
 The `ScheduleDef` built-in (`internal/tools/builtin/scheduledef.go`) is a 7-op tool — the five core ops `create` / `fork` / `get` / `list` / `retire` plus `add_hook` / `remove_hook` (each hook edit persists a new lineage version). Static yaml entries remain immutable ground truth; the tool authors *new* names only.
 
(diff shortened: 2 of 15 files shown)


denn-gubsky:
**Self-review pass** — three findings, all fixed in `fix(scheduler): a channel write honours the channel's declared retention`.

1. **A cadence writer ignored the operator's retention.** Both scheduler publish paths passed `maxMessages=0` and no expiry. For a *tick every minute* that is ~500k rows a year on a channel whose operator did set `default_ttl`/`max_messages` — the limits just never reached the writer, so nothing trimmed and the TTL sweeper had nothing to reap. `ChannelScopeResolver` now returns the channel's declared shape (`DeclaredChannel`), and both writers apply it; the `on_complete` hook path had the same gap at run cadence and is fixed by the same seam.
2. **The consolidation fan-out branch came first.** It keys off a metadata flag, and on a channel tick `metadata` is opaque *payload* — so a tick whose payload carried that key would have dispatched a wave of runs, the exact opposite of "no run". Delivery is decided first now.
3. **An unknown delivery fell through to the run path** in `unmarshalDef`, firing the agent on the least-understood input. Refused at decode now (both write paths already refuse it; this is the fail-closed backstop).

`go test ./...` still clean (100 packages), plus two new scheduler tests.

denn-gubsky:
**Verified live**, on a throwaway instance (`:8912`, mock provider, its own data dir — the `:8799` memory-test instance untouched):

```yaml
scheduled_runs:
  minute-tick:
    delivery: channel
    channel: wave-in
    schedule: "* * * * *"
    metadata: { batch: nightly }
channels:
  wave-in: { scope: global, max_messages: 10, default_ttl: 3600 }
```

- the static schedule **materialized into the substrate** with `Delivery`/`Channel` round-tripping (`/v1/_schedules/list-all` shows both);
- the tick landed on the channel:
  ```json
  {"schedule_name":"minute-tick","fired_at":"2026-09-11T04:03:18.509529Z",
   "delivery":"channel","payload":{"batch":"nightly"}}
  ```
- **`runs` table: 0 rows.** The tick started no run — the whole point of the mode, and the one thing the unit tests assert by proxy;
- `schedule_run_state`: `last_status=completed`, `fire_count=1`, no error — a tick is a fire;
- the message carries **`expires_at = published_at + 3599.998s`**, the channel's declared `default_ttl` — the retention fix in the review commit, working through main.go's resolver closure, which is the one line no test reaches.

PR #1183, opened 2026-09-11
feat(client): the TeamDef version lifecycle reaches the SDK

denn-gubsky:
**RFC CY item 18.** `list`, `promote`, `retire` and `verify` have existed on the TeamDef substrate since it shipped and are reachable over HTTP, but `LoomcycleClient` exposed only create / fork / get / delete / run / diagram — so **a client that could author a workflow could not put one in force.**

That gap has a sharp edge. `forkTeam` defaults to `promote:false`, *deliberately*: an edited graph should be reviewable, diagrammable, even runnable by `def_id` before it becomes what the **name** means. Without `promoteTeam` the second half of that two-step was unreachable, so a consumer either hand-rolled the POST or gave up the review step by forking straight into force.

```ts
listTeamVersions(name)            // every version of one team, newest first
promoteTeam(defId)                // point the active pointer
retireTeam(defId, retired)        // soft-retire, and un-retire
verifyTeam(name, contentSha256)   // drift check against the deployed version
```

`verifyTeam` is the one an operator needs and couldn't have: a workflow kept in source control and pushed to several deployments had no way to ask *"is what I have what is running"*. An absent team **answers** `deployed: false` rather than raising — which a caller must distinguish from a deployed version whose hash differs. The first is "never shipped"; the second is drift.

**No server change** — these are the wire shapes the substrate already returns.

## Tested

The TS suite: **296 tests green**, 18 in `teams.test.ts`. Each new method asserts the URL, the method, and the snake_case body the runtime expects, plus two cases that are easy to get wrong:

- `retired: false` must reach the wire **as false** rather than being dropped as a falsy value — otherwise un-retiring is a no-op that reports success;
- a name with no active version must come back as an **answer**, not an error.


denn-gubsky:
**Release note for whoever merges this:** the adapter's `package.json` stays at **1.72.0**, so this change does *not* publish on its own. `@loomcycle/client` publishes only when the package version equals the release tag, and the adapter has been behind the binary (v1.76.0) for four releases — so these methods, like the `PathToolInput` pagination types already in-tree, sit unpublished until the adapter version is bumped to match the next tag.

Deliberately not bumped here: picking a version is a release decision, not a feature one, and guessing the next tag would either publish early or skip again.

PR #1186, opened 2026-09-11
fix(channels): every writer that resolves a channel def honours its hold

denn-gubsky:
#1180 shipped the hold with the claim that putting the check in `StorePublisher` makes every internal publish path honour it — *"a promise enforced at five call sites is a promise the sixth one breaks."* Then #1181 added a sixth writer that doesn't go through the publisher, and it broke exactly that way.

## How it was found

By running the two features **together**, which is the pairing they exist for: a cron tick into a `hold: true` channel — cadence arrives, the workflow waits for a human.

```
peek after the tick → {"messages":[ …the tick… ]}     # delivered immediately
release             → {"released":[],"still_held":0}  # nothing was ever held
```

The scheduler writes via `store.ChannelPublish` directly, so `HoldFn` never ran. Neither unit tests nor either PR's own live check could see it: each feature was correct alone.

## The fix

Four writers reach the store without the publisher. **Two resolve the channel definition** and therefore owe it obedience:

- the scheduler's `delivery: channel` tick (the regression), and
- the scheduler's `on_complete: channel.publish` — same gap, older than this line.

Both already learn scope and retention from `ChannelScopeResolver`, so hold rides the same lookup: `DeclaredChannel` gains `Hold`, and a held target writes the reserved `visible_at`.

The **webhook's `on_complete: channel.publish`** is the third. It resolved nothing, so `lookup.WebhookStore` gains `ChannelGet` (`store.Store` already satisfies it) and the receiver gets the yaml-then-runtime resolution `(*http.Server).ChannelHeld` uses.

**The rule, now true: a writer either resolves the channel definition or honours nothing from it.** The webhook hook still doesn't consult `max_messages`, TTL or the declared scope — widening *that* moves where existing messages land, which is a separate change with its own blast radius. A hold is not: honouring it can only withhold a message the operator asked to be withheld.

The fourth writer is Document's change-event ring, which publishes to a **synthetic** name (`documents/<id>/chunks`) no operator can declare. An undeclared channel has no definition to honour, so there's nothing to fix.

## Tested

`go test ./...` clean (**100 packages**, full output grepped for `FAIL`).

- **scheduler** — a tick to a held channel delivers nothing *and is releasable* (stored, not lost); an `on_complete` hook likewise;
- **webhook** — the hook honours a hold from operator yaml **and** from a runtime-declared row, with an unheld channel as the control.

**Verified live**, the same way the bug was found: the cron tick now peeks **empty** after firing, and one release hands over the cadence signal with `still_held: 0`.

### DIFF
--- a/cmd/loomcycle/main.go
+++ b/cmd/loomcycle/main.go
@@ -2945,6 +2945,7 @@ func main() {
 				Scope:       def.Scope,
 				DefaultTTL:  def.DefaultTTL,
 				MaxMessages: def.MaxMessages,
+				Hold:        def.Hold,
 			}, true
 		})
 		// RFC BL P2 consolidation fan-out: the provider resolver decides
--- a/internal/api/webhook/oncomplete.go
+++ b/internal/api/webhook/oncomplete.go
@@ -111,9 +111,14 @@ func (rec *Receiver) dispatchOnCompleteChannelPublish(ctx context.Context, name,
 // (operator-global, so every tenant sees it), then the tenant's runtime row.
 //
 // A store fault answers false: an unreachable definition plane must not start
-// holding a channel nobody declared held. Mirrors (*http.Server).ChannelHeld,
-// which the in-process publishers use; the receiver has its own because it
-// holds a narrow store interface rather than the server.
+// holding channels nobody declared held. The choice is nearly moot in practice
+// — the publish on the next line uses the SAME store, so a fault here means the
+// message is not written either — and "don't invent a hold" is the safer
+// default for a transient error. A yaml-declared hold never touches the store
+// at all.
+//
+// Mirrors (*http.Server).ChannelHeld, which the in-process publishers use; the
+// receiver has its own because it holds a narrow store interface, not a server.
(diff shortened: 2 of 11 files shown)


denn-gubsky:
Two follow-ups on this branch after a cold re-read.

**The read-side twin of the census.** The writer census asked "who can write past a hold"; the matching question is "who can silently *release* one". Only one path rewrites `visible_at` from a caller-supplied value — `SnapshotRestoreChannelMessage` — and while it preserves the instant today, `ChannelPublish` *clamps* a past `visible_at` to now, so a future edit that normalised the restore path the same way would open every breakpoint in a restored deployment. Now pinned on both backends.

For the record, the rest of the census: of 14 SELECTs against `channel_messages`, the two **delivery** paths (`channelRead`'s two arms, backing subscribe / peek / await) both filter `visible_at <= now`. The others are the snapshot export, the overflow trim's keep-set, `ChannelRelease`'s own two queries, and the stats aggregate — none deliver, and each wants every row. No read path leaks a held message.

**A doc fix:** inserting `ChannelGet` left `lookup.WebhookStore`'s own comment saying the receiver "needs only these two".

PR #1248, opened 2026-09-15
feat(chat): on-the-fly context distillation for the bundled chat agents

denn-gubsky:
The three bundled chat agents relied on `compaction:` alone — a prose summary fired at 80/70% of the window. They now also carry a `context:` block, so a long conversation is distilled **continuously** rather than only at a threshold.

```yaml
context:
  mode: auto
  keep_last_n: 6
  reasoning: recap
  autorecap_at_pct: 80        # 70 on the two local agents
```

## `mode: auto`, not a pinned mode — that is the substance

The ask was structured-JSON compaction: L2 stateful, where a JSON state Σ replaces the conversation entirely. **That is not available to these agents**, and the source says so directly (`internal/loop/stateful.go:28`):

> *"PR1 scope: autonomous runs. Interactive steering, pause/park, and cross-instance resume of a stateful run are not wired here yet."*

All three chat agents are interactive — `unbounded_iterations`, park at `end_turn`, steered from the `/run` terminal. That is what they are for.

Two details make this load-bearing rather than pedantic:

- **The runtime already refuses the pairing itself.** `resolveAutoContextMode` routes any interactive *or* local run to `recap`, "regardless of tier", citing exactly this limitation.
- **But an explicit `mode: stateful` has no guard.** `contextStatefulMode` branches unconditionally into `runStateful`. Pinning it here would have taken the stateful loop and silently cost these agents steering and parking — the terminal flow they exist to serve.

So `auto` is not a hedge. It resolves to recap today, and picks up Σ **by itself** the day steer/park compose onto that loop — no bundle edit, nobody needing to remember. Pinning `recap` would behave identically now and quietly never improve.

There is a second-order reason too: in stateful mode the model is offered exactly one tool (`emit_state`) and reaches its real 16 tools one action per step via an `action` field. That is a different interaction shape from conversational chat, independent of the steer/park gap.

## The rest

`keep_last_n: 6` keeps the recent exchange verbatim, snapped to a clean user-turn boundary so a `tool_use`/`tool_result` pair is never split. `autorecap_at_pct` mirrors each agent's existing compaction trigger — 80 for `chat/medium`, 70 for the two local agents whose windows fill sooner.

`compaction:` is left in place: recap takes precedence over its auto-trigger, and `memory_flush: true` still does its job of banking discarded turns onto the consolidation queue.

## Verified

- Edited **both** copies — `bundles/chat/loomcycle.yaml` and the embedded `cmd/loomcycle/embedded/bundles/chat.yaml` — which a sync test compares byte-for-byte. Probed it by desyncing them: `TestChatBundle_SourceMatchesEmbedded` fails as it should.
- `LOOMCYCLE_PRESETS=base,chat … validate` → OK, all three agents resolve.
- **Confirmed the block is parsed, not silently discarded**: loaded the bundle and read `Agent.Context` back per agent — `mode=auto keep_last_n=6 reasoning=recap`, autorecap 80/70/70. `validate` passing only proves nothing *rejected* it, and this repo has shipped silent-config-discard bugs before.

Targeted suites (`cmd/loomcycle` bundle + preset tests, `internal/config`) all pass. The full `-race` run was at 92/103 packages with **0 failures** when this was opened; I will confirm the final number on the PR rather than leave it implied.


denn-gubsky:
Full suite confirmed since opening: `go test -race -timeout 1800s ./...` — **0 FAIL, 103 packages, 0 data races**. (The body noted 92/103 mid-run; this is the final number.)

PR #1250, opened 2026-09-15
feat(prompt): the injected tool list says which tool to reach for instead

denn-gubsky:
Three asks. **Two needed no runtime change** — worth saying up front, because the interesting part is which one did and why.

## 1. The chat agents read the user root, not only the tenant root

They injected `{{memory:tenant_info}}` — what is true of the deployment — and nothing about the person they were talking to. All three now also inject `{{memory:user_info}}`, which composes the user-scope `/memory/user_root` document with the `human` core block.

No runtime work: `user_info` is already a known variant and `renderUserInfo` already reads the user-root document. The comment on `VariantUserInfo` still says *"the user-root document is a later phase"*, which stopped being true when `readUserRootMarkdown` landed — the same stale-comment shape `VariantTenantInfo` carried before it. Left alone rather than fixed in passing.

Both placeholders are gated on an actual reference because rendering either **provisions** its root document; adding `user_info` here is what makes the user root exist for these agents. All three already hold `memory_scopes: [user]`.

## 2. `{{tool:Context.tools}}` was already narrowed — verified, not assumed

`renderContextTools` passes `mi.Tools` — the run's already-resolved `allowedTools` — as **both** the candidate set and the ctx allowlist, so `op=tools`' intersection is a no-op on an already-filtered list. The inventory is what the agent can actually call, narrowed by construction rather than by filter. No change needed.

## 3. The guidance already existed; the injection was dropping it

This is the one that was real, and the cause is not what it looks like.

Every tool description already states what the tool is **not** for and which neighbour owns that job — `"Do NOT use it to match file names — that is Glob."` I put those there in #1228. But `formatToolLine` rendered only `firstSentence(desc, 120)`, and the boundary lives in a *later* sentence, so it was truncated away.

So each bundle hand-wrote a "Choosing among them" section restating what the descriptions already said — prose that drifts the moment a tool changes, and that only the bundles carrying it ever benefit from. The data was there; the pipe was too narrow.

The inventory now renders the clause verbatim on a continuation line:

```
- Grep (read) — Search file CONTENTS on a volume with an RE2 regular expression.
  Do NOT use it to match file names — that is Glob.
- WebFetch (read) — GET one http(s) URL and return its body with HTML stripped to readable text.
  Do NOT use it to FIND pages: use WebSearch to discover URLs first.
```

**Verbatim rather than reformatted into a house style.** The clause already names the neighbour; paraphrasing would be this layer inventing guidance the tool author did not write.

`"do not use"` is a reliable anchor rather than a heuristic: **both** description rubrics — `description_rubric_test.go` for in-band tools and `tools_surface_test.go` for the MCP surface — require every description to contain it. A tool cannot ship without one, so extraction cannot silently degrade into finding nothing.

Two degradations are deliberate: a description with **no** boundary renders exactly as before, and a clause too long to trust to truncation is **dropped rather than clipped** — half a sentence naming half a tool is worse than no guidance.

### The chat bundle

Drops its three hand-written sections (**−64 lines, +33**) and adds `{{tool:Context.guide}}` for the how-to-call half (op enum, required args, usage hints).

**Two bespoke traps are kept rather than silently lost:** that `Memory op=recall` is the only semantic search over remembered facts, and that `Document graph_recall` returns `seeds: 0` on a store with no entity metadata. Neither appears in any description or usage hint, so deleting them would have been a regression dressed as cleanup. They belong in `UsageHint` on Memory/Document — a follow-up, not a bundle edit.

## Blast radius, stated

The boundary now lands in **every** agent using `{{tool:Context.tools}}`, not just chat. That is the intent — the guidance stops being per-bundle prose anyone can forget to update — but it does add one line per tool to those prompts.

## Testing

Targeted tests pass: boundary extraction against the real `Grep`/`Glob`/`WebFetch`/`Read` descriptions, verbatim-not-paraphrased, absent-or-oversized yields nothing, and a no-boundary description renders byte-identically to before. Bundle sync + preset tests pass; `LOOMCYCLE_PRESETS=base,chat … validate` → OK on all three agents.

Confirmed fail-before: removing the render line reports *"inventory line does not name the neighbour"* for Grep and Glob.

The full `-race` run was still early when this was opened; I will post the final number on the PR rather than leave it implied.


denn-gubsky:
Full suite confirmed since opening: `go test -race -timeout 1800s ./...` — **0 FAIL, 103 packages, 0 data races**. (The body noted the run was still early; this is the final number.)

PR #1253, opened 2026-09-16
fix(resume): a paused run comes back on its own configuration, not its definition's

denn-gubsky:
This is the snapshot/resume fidelity core: a resumed run comes back as the run
that paused, or it says why it cannot. It also carries the model-restore commit
(`3127b8c`, `28859aa`) that was on this branch already.

## Why

A paused run came back configured from its **agent definition**, not from itself.

Every call-time value a run started with — temperature, compaction policy,
layered-context mode, context window, run timeout — was rebuilt from the def on
resume. Four lines in `resume.go` said so out loud:

```go
RunTimeoutSeconds:   agentDef.RunTimeoutSeconds, // per-run override not snapshotted
Sampling:            agentDef.Sampling,          // per-run override not snapshotted
Compaction:          agentDef.Compaction,        // per-run override not snapshotted
Context:             agentDef.Context,           // per-run override not snapshotted
```

A long interactive chat that paused and resumed silently reverted its own
settings mid-conversation, and nothing in the transcript said it had changed.
Honesty in a comment is not the same as correctness.

The caller's `allowed_hosts` narrowing was dropped the same way, with one
difference that matters:

```go
loopCtx = tools.WithHostPolicy(loopCtx, tools.HostPolicyValue{}) // no caller narrowing snapshotted
```

Every other lost override merely changed behaviour. That one handed the resumed
run the **bare operator floor** back — the only case where losing an override
moved a boundary outward. Contained (the floor still bounds it), but "resuming a
run cannot increase its reach" is worth actually holding.

## What

These are properties of the **run**, not of the definition it started from, so
they now live on the run.

- a `run_config` column on `runs` (sqlite + postgres `0078`), **opaque JSON**:
  `internal/config` imports `internal/store`, so the store cannot name these
  types — and has no business interpreting them. It persists bytes.
- `PausedRunEntry.RunConfig` carries the record through the snapshot envelope,
  so it survives a move between deployments.
- resume restores it and falls back to the definition only when there is no
  record — which is exactly what every run did before, so legacy rows resume
  unchanged.

Each of the four run-creation paths (`RunOnce`, `handleRuns`, `handleMessages`,
`prepareSubRunValues`) merges the record **once** and hands the same values to
both the run row and the loop, so the durable record cannot drift from what the
run actually ran with. In `RunOnce` and `handleRuns` this removed a merge that
was already being computed twice.

**The narrowing MODE is deliberately not restored.** Whether a caller's list
replaces the operator floor or intersects with it is operator config, re-read at
resume — the same call a fresh run with the same inputs would get. Only the
caller's list and filter come from the run. `has_list` is carried explicitly
because an empty list means *allow nothing*, a distinction `omitempty` erases.

### Three adjacent gaps closed, each the same defect wearing a different hat

- resume never stamped the layered-context policy on the loop ctx, so a resumed
  run's **sub-agents** inherited a different retention mode than the original's
  children. Every non-resume path stamps it.
- the Recall grant was re-derived from the def, so a run that enabled recall
  per-run lost the Recall tool on resume.
- **`POST /v1/sessions/{id}/messages` never recorded that a run was
  interactive** (own commit, `1b0c62f`). It accepts `interactive`, passes it to
  the loop, and no write sets the column. An interactive run parks at end_turn;
  resume reads `runs.interactive` to choose park-vs-complete. So a conversation
  continued through the terminal's own continuation path came back from a
  snapshot as a **batch run** — run to completion, session over, instead of
  parking for the next operator turn. The column was already captured and
  restored correctly; only the write was missing. Split out so it can be dropped
  independently if you'd rather it went on its own.

## Deliberately out of scope

Restoring a mid-run **fallback target**. `runs.model` is written once at
`CreateRun`; a fallback emits `EventProviderFallback` without updating the row.
So this restores the run's own *starting* model, not where it failed over to.
Recording fallbacks on the row is a hot-path write and a decision in its own
right.

`max_tokens` and `max_iterations` stay def-derived on resume because they are
not per-run overridable today.

## What was tested

The record is **threaded**, not computed — merged at run start, copied onto the
run row, read back on resume, decoded, handed to the loop. Testing either end
proves nothing: delete the carry and both stay green. So every test starts
upstream and asserts on the **consumer's** view.

| test | crosses | asserts on |
|---|---|---|
| `TestResumedRun_KeepsItsOwnSampling` | run start → row → resume → loop | the provider Request's temperature |
| `TestResumedRun_KeepsItsContextWindow` | same | the provider Request's context window |
| `TestResumedRun_KeepsItsCallerHostNarrowing` | same, plus the tool set | whether the resumed run's HTTP tool can actually reach the host, with an **empty operator floor** so a dropped narrowing reaches nothing |
| `TestRoundTrip_PreservesRunConfig` | capture → envelope → restore | the captured JSON, not just the store |
| `TestStoreContract/CreateRunConfigRoundTrip` | both backends | present **and absent** — a run with no record must read back NULL, not `{}` |
| `TestMessages_RecordsThatTheRunIsInteractive` | the continuation path → the row | the column resume reads |
| `TestRunConfigCensus_…` | — | a **writer census**: any function building both a `store.RunIdentity` and a `loop.RunOptions` must persist the record |
| `TestSubRunConfig_DefRewritesLeaveTuningUntouched` | — | the one assumption the sub-agent path rests on |

Each fixture makes *restored* and *re-derived from the definition* produce
different values (0.11 vs 0.9, 65536 vs 8192), so no assertion can pass while
restoring nothing.

**Eight fail-before probes, all confirmed**, each reverting one carry:

1. resume re-derives sampling + window from the def → both loop tests fail
2. resume drops the host narrowing → the run cannot reach its own host
3. `handleRuns` stops persisting the record → all three loop tests **and** the census fail
4. the sqlite write is dropped → contract + snapshot fail
5. the snapshot **capture** carry is dropped → the envelope has no record
6. the snapshot **restore** carry is dropped → capture stays green, restore fails
7. the postgres write is dropped → contract fails on PG (sqlite would stay green)
8. postgres writes it but the scanner never assigns it → the read seam fails

The census also refuses to pass if it matches fewer than four sites — a census
that matches nothing passes forever.

Verified against a **live Postgres**, not just sqlite: the full
`./internal/store/postgres/` suite green (232s), which is what exercises the
JSONB `::text` read cast and the `::jsonb` write cast that sqlite cannot.

**Full `-race` sweep: in progress at the time of opening.** The host ran out of
memory partway through two earlier attempts (the runner was killed, no test
failed), so it is re-running with package parallelism capped at 2 — same
`-race -timeout 1800s`, only fewer packages built concurrently. I'll post the
result on this PR. Green so far, and CI runs it independently.

Already green on this exact tree, package by package:
`./internal/api/http/`, `./internal/snapshot/`, `./internal/store/...`,
plus the live-Postgres run above.

---


denn-gubsky:
Full `-race` sweep finished clean on the merged tree:

```
go test -race -timeout 1800s -p 2 ./...
FAIL: 0   ok: 103   races: 0   (exit 0)
```

`-p 2` only caps how many packages build concurrently — the host OOM-killed two earlier attempts mid-run, which is why it was still in flight when this opened. No test ever failed in those runs; they were killed, not red.

PR #1268, opened 2026-09-17
fix(memory): the traces selector, backfill targeting, and the silent grants that hid them

denn-gubsky:
## Why

A LoCoMo re-baseline measured an answerer at **0.0000 accuracy** on a store that
was answering fine. Chasing that number down surfaced three defects that share
one shape: **a selector or grant that fails silently, and degrades to something
that looks like a result.** None of them raised an error, appeared in a log, or
produced an empty response distinguishable from a genuine miss.

The benchmark work that found them is included because the config *is* the
reproduction — but the two runtime commits stand on their own and can be split
out if you'd rather review them separately.

### 1. `POST /v1/_memory/search` silently ignored `sources:["traces"]` (runtime)

The handler kept its **own copy** of the wire→enum mapping, and that copy had no
case for `traces`. The value was dropped, the selector collapsed to empty, and
`SearchQuery.Filter` read empty as "no selector" — which for traces is the
*active exclusion* branch. **Asking for raw turns returned everything except raw
turns**, with a 200 and no warning.

The same duplicate had already drifted once in the other direction (`notes` was
missing from the in-band tool's copy). The file holding both even diagnoses the
cause, for a sibling field:

> *"Two hand-maintained copies of a predicate that decides which rows get DROPPED
> is how this endpoint and its in-band twin have already drifted twice; one
> struct and one parser is the fix."*

That was applied to `when` and not to `sources`. So: `memrank.ParseSources` moves
into `internal/memory` beside `Source`/`SearchQuery`/`Filter`, both copies are
deleted, and drift becomes structurally impossible rather than tested for.

**It also stops failing open.** `sources` *narrows*, so an empty selector means
"the default" — a caller who misspells one name doesn't get a smaller result set,
it gets the **default** one while believing it filtered. An unknown value
*alongside* a known one is still dropped (a newer caller's additive value must not
break an older runtime); an unknown value *alone* is now refused.

⚠️ **Behaviour change worth a release note:** a caller sending only a misspelt
source used to get a silent default and now gets a 400.

Wiring traces through also makes `ErrTracesNotCombinable` reachable for the first
time from either surface, and neither mapped it — it would have been a blind 500.
Mapped beside the existing refusal.

### 2. `backfill_traces` could bound its work but not choose its subject (runtime)

`limit` caps how many chats are examined and `ListSessions` orders newest-first,
so a bounded run walks into the operator's most **recent** chats — the opposite of
"make my archive searchable", which is the reason the endpoint exists.

Concretely, on the benchmark store: `limit=500` scanned 500 sessions, hit
`stop_reason: limit`, and reached **zero** of the 19 real conversations. Worse,
1608 of that user's 1627 sessions were evaluation runs whose prompts quote the
answer key — **409 turns containing the literal string `Gold answer`** — so an
unfiltered backfill would have embedded the answers into the index the evaluation
then searches.

`store.SessionFilter` already carries `AgentName`; the handler was discarding it.
This exposes `?agent=` and stops throwing the field away. Optional — omitted, the
whole-user sweep is unchanged.

### 3. A tool grant is not a read grant (bench config)

`tools: [History]` puts History on the agent but does **not** grant the read;
`history_scope` is a separate default-deny policy. An answerer holding History
made **123 window/search calls and had all 123 refused** while the run looked
entirely healthy — the model absorbs the tool error and answers from what it
already had. The accuracy it reported was real but described an agent that had
never read a single original turn.

Swept the repo: 4 defs grant History, all now carry the read grant
(`memory/consolidator` already did), and no def with `tools` unset lacks it.
**The shipped bundles were never affected** — this was confined to the bench defs.

## What this unblocked

With all three fixed, the P-A retrieval comparison ran cleanly — same frozen
263-fact store, same 150 questions, same model and judge, arms differing in one
source and one prompt bullet:

| arm | graded | accuracy | strict | abstention |
|---|---|---|---|---|
| control (facts only) | 149 | 0.3960 | 0.315 | 26.8% |
| window (fact → its session → turns) | 148 | 0.5473 | 0.459 | 16.9% |
| **traces (question → turn directly)** | 146 | **0.7877** | **0.719** | **6.2%** |

Exact McNemar on strict correctness: control→window **+24/−3, p=4.9e-05**;
control→traces **+61/−3, p=4.7e-15**; window→traces **+45/−8, p=2.4e-07**. The
trace route is the first configuration to clear the proposed gate
(accuracy ≥ 0.67 and strict ≥ 0.60), and it independently replicates the earlier
"turns win 0.738 vs 0.157" result on a different rig with the fixed extractor.

## What was tested

**⚠️ The drift guard itself contained the bug it guarded against.** The old
`TestParseSources_MatchesTheHTTPParser` asserted over a hand-written
`{facts, notes, documents}` and its comment claimed *"adding a fourth kind fails
here until both are updated"* — it did not, because that list was a **third** copy
of the vocabulary. Every assertion now derives the enum from the tool's advertised
input schema, so a fifth source that isn't wired fails CI.

Fail-before verified rather than assumed:

- Drop `SourceTraces` from the parser → `TestParseSources_AcceptsEverySourceInItsOwnSchema`
  and `_IsTheOnlyCopy` FAIL.
- Drop the refusal → `_UnknownAloneIsRefusedButMixedIsDropped` FAILs with
  *"must be refused, not silently widened to the op default"*.
- Remove the agent filter → `TestBackfillTraces_AgentSelectsWhichChats` FAILs with
  *"scanned 2 sessions, want only the 1 served by corpus-scribe"*.
- `TestBackfillTraces_NoAgentStillScansEverything` is a no-regression pin and
  passes with or without the change, by construction.

End-to-end against a live server:

| probe | before | after |
|---|---|---|
| `sources:["traces"]` | returned facts, 200 | `kind=trace` rows |
| `sources:["tracez"]` | silent default, 200 | 400 naming valid values |
| `sources:["traces","facts"]` | unreachable → blind 500 | 400, combinability refusal |
| `sources:["facts","future_kind"]` | filters to facts | unchanged |
| no selector | every plane, traces excluded | unchanged |

That last row is the RFC CI requirement that an unfiltered search stay
byte-identical to what it returned before traces existed.

`go vet` clean. `internal/memory`, `internal/tools/builtin` and `internal/api/http`
all pass under `-race` on the rebased tree.

⚠️ **Two failures in the full-suite run, and they are NOT from this change — but
they are not "flake" either.** `TestConsolidator_BackfillNeverCrowdsOutThisPasses
Facts` and `TestConsolidator_QueueDrainsAcrossSeveralExtractorCalls` both failed
with `code_agent_timeout: run exceeded its 30s wall-clock budget`. What was
checked:

- The failure is a **timeout, not an error** — no `ErrSourcesUnrecognised`
  anywhere, and the `Memory.recall` calls in both traces plainly succeeded.
- **The consolidator never passes a `sources` selector at all** — all three
  matches in `bundles/memory/loomcycle.yaml` are comments — so this change
  cannot reach that code path.
- Both tests **pass in isolation under `-race` at `origin/main`** (38.6s) **and
  on this branch** (52.0s).

So they are a **30s code-agent budget that is load-sensitive under parallel
`-race`** — the run happened on a machine concurrently hosting a loomcycle
benchmark server and Postgres. Worth knowing for CI capacity; flagging rather
than burying it, since "passes in isolation" is not the same as "passes in CI".


denn-gubsky:
Closing the loop on the two `TestConsolidator_*` timeouts flagged in the description.

Re-ran `go test -race -timeout 1800s ./cmd/loomcycle/` on a quiet machine (bench server and Postgres stopped):

```
ok  github.com/denn-gubsky/loomcycle/cmd/loomcycle  159.018s
```

Green. Under the loaded full-suite run the same package took **306s and failed those two tests** on the 30s code-agent wall-clock budget; quiet it takes **159s and passes**. So the characterisation in the description holds — load-sensitive budget, not a defect, and not caused by this change (the consolidator never passes a `sources` selector; both tests pass in isolation at `origin/main` too).

Worth noting it lands next to #1269, which gave the Postgres store contract headroom over its own 30-minute cap — same underlying theme of time budgets being tight under parallel CI load. If those two consolidator tests start flaking in CI, the budget is the knob, not the test.

PR #1271, opened 2026-09-17
fix(document): bound graph_recall, and provision a scope's schema once per process

denn-gubsky:
## Why

Two reviews of the memory epic's implementation, one commit each.

# Commit 1 — RFC CV P5 bounds

A review of **RFC CV P5** (`40b6ced3`) — the newest code in the memory epic and the
capstone the RFC DB gate existed to unblock — turned up five issues. Two are
bounded-resource defects on a path an agent invokes once per question, and one of
those has a **silent** degradation mode.

Scope note: this reviewed `internal/tools/builtin/document_graph.go` (689 lines).
The epic is ~44k lines; the rest is not covered by this PR.

### 1. A hop could return unbounded rows, and the fallout was silent

`graphFrontierCap` bounds how many nodes a hop expands **from**. It never bounded
how many rows a hop **returns**, which is nodes × degree — and degree is not small.
On a 304-chunk benchmark store the maximum in-degree is **162**: one hub entity with
162 facts pointing at it. A single frontier node returns 162 rows; a full 500-node
frontier returns five figures, all materialised.

The knock-on is worse than the memory. Those rows accumulate into `order`, which
`graphIdentityNodes` turns into **one placeholder per id**. Past the driver ceiling
(SQLite 32766, Postgres 65535) that query fails — and it fails *silently by design*,
returning an empty set — so assertion-first budget ordering reverts to the exact
behaviour it was added to fix, on precisely the dense graphs where it matters most.
The existing "a miss is safe in the direction that matters" argument holds for a
store error, but not here, where the failure is **caused by success**.

Adds `graphHopRowCap` (a `LIMIT`, fetched one over as a truncation sentinel) and
`batchIDs` on every `IN (...)`, and reports the cap through the existing
`truncated` flag.

### 2. The backfill was an N+1 that queried before it checked the budget

One query per ranked id, up to `graphSeedTopK` of them — and the budget test ran
*after* the fetch, so a walk whose budget was already full still paid for every
remaining round trip and threw each result away. `graphIdentityNodes` two hundred
lines earlier does the right thing ("One round trip for the whole set"); this is the
same shape. Now one batched read, walked in **rank order** (the batch returns rows
in whatever order SQL likes).

### 3. `limit` was silently ignored whenever `budget_chars` was set

It sat in the `else` of the budget test, and the backfill ignored it too — so a
caller passing both got the row cap accepted and dropped, with no signal that it had
not applied. That is the same shape as the `sources` selector that decoded and was
discarded (#1268), and as `limit` vs `top_k` on the search path before it. The
bounds now compose: budget caps by content, limit caps by rows, whichever binds
first wins.

### 4. `graphSeeds` was dead code

P5 replaced it with `graphSeedIDs` and left the wrapper. Deleted, per the convention
against keeping stubs.

### 5. The file header argued for a cap the code no longer has

> *"one round trip per hop, capped at two hops. Two is not arbitrary…"*

`graphMaxHops` is 6, and the constant's own comment explains why it changed.
Rewritten to describe the three axes that actually bound a walk.

## What was tested

- **`TestGraphNeighbours_CapsRowsPerHop`** — builds a hub one past the cap; asserts
  the row count is bounded and that the hop reports hitting it.

  ⚠️ **It guards the contract, not the `LIMIT` specifically.** The bound has two
  layers (the `LIMIT` with its sentinel, and the Go trim that removes the sentinel),
  and removing only the `LIMIT` leaves the trim still satisfying the assertion — my
  first fail-before passed for exactly that reason. Verified to fail with **both**
  removed: *"one hop returned 2001 rows with a cap of 2000"*. The `LIMIT` is what
  stops the database materialising and transferring those rows, which is not
  observable from the tool's output; it is defence at the layer that can act on it
  rather than a separately testable claim.

- **`TestGraphRecall_LimitBoundsEvenWithABudget`** — the facts arrive by
  **expansion**, not as seeds: a seed list is itself bounded by `limit`, so seeding
  would test the wrong query. My first fixture made that mistake and proved nothing.
  Fail-before: *"returned 13 chunks under limit=3 with a budget set"*.

- **`TestBatchIDs_SplitsAtTheCeiling`** — ⚠️ the only layer at which the
  placeholder-ceiling fix is testable: the end-to-end failure needs more ids in one
  `IN (...)` than the driver accepts, and a fixture that large is not a unit test.
  The split is pinned here and the call sites go through it. Fail-before:
  *"1 batches, want 3"*.

- **The existing `TestGraphRecall_*` suite passes unchanged**, which is what guards
  the behaviour P5 was measured and shipped on.

`go vet` clean. `internal/tools/builtin` passes in 28s; a `-race` run at the full
suite's own `-timeout 1800s` is reported in a comment below.

⚠️ **Test-cost note, because this package is close to a real ceiling.** Under `-race`
`internal/tools/builtin` already takes **1475s (24.6 min)** on `main` — that is a
pre-existing baseline, not something this PR changes — against a 30-minute cap that
#1269 just had to raise headroom for elsewhere. The first version of
`TestGraphNeighbours_CapsRowsPerHop` built a fixture one past the production cap
(2001 chunks, 2001 edges) and cost **95s under `-race`** on its own. Making the cap a
parameter brought that to **1.2s** for a strictly stronger assertion. Worth knowing
before adding fixtures sized to production constants in this package.

## Not fixed here — a behaviour question, not a defect

Explicit `seed_ids` are silently truncated by `limit` in the seed query
(`fetch := limit`): hand in 12 ids with `limit: 3` and you get 3 seeds, no signal.
Arguably intended, since `limit` bounds results — but it is the same
accept-then-drop shape as #3, and it is what made my first test fixture mislead me.
Worth deciding deliberately rather than inheriting.

---

# Commit 2 — `document.go`

A review of `document.go` (4909 lines) found four more. The first is the substantive one.

### 1. `ensureSchema` ran in full on **every** op

Sixteen `CREATE … IF NOT EXISTS` statements, an asset DDL and nine migration probes
ahead of the actual work — on `get_chunk`, `graph_recall`, `list_facts`, everything.

Measured controlled: both arms in one process, interleaved across rounds, with the
"before" arm reproduced by dropping the memo ahead of each call (which makes
`ensureSchema` re-provision — exactly the old behaviour):

| | per `get_chunk` |
|---|---|
| before — provision per call | **452.96 µs** |
| after — memoised | **110.83 µs** |
| removed | **342 µs — 4.1× faster; schema was 76% of the call** |

The before arm lands within 0.1 µs of an independent measurement taken before any of
this was written (453.08 µs), which is the check that the A/B reproduces the real
thing. Provisioning alone is 334.9 µs against a memoised 53 ns — a factor of 6295.
That is the **sqlite** tier; on postgres each of those statements is a network round
trip. One benchmark run issued 915 Document calls, so ~24k round trips existed only
to re-assert a schema that was already there.

The memo lives on the `sqlmem.Manager`, not on `*Document`: `&Document{}` is built
ad-hoc in about eight places (the Memory tool's placement helpers, `Context`, the
HTTP describe handler) and they all share one Manager pointer, so a per-struct cache
would miss every one of them.

**⚠️ A memo is a claim about the process, not the database, and the dangerous case is
a missing COLUMN rather than a missing table.** `CREATE TABLE IF NOT EXISTS` leaves
an existing table alone, so a column a later release adds arrives only through the
`ALTER` in `provisionSchema` — which now runs once per process. If anything leaves a
scope on the old shape after that, every write of the new column would fail
**forever**. So a statement failing on a missing column (or table, or schema)
re-provisions and retries once: the memo self-heals, and a pending migration costs a
slow call instead of a dead scope.

**⚠️ And that retry is infinitely recursive without a guard**, in a way that does not
announce itself. The migrations ask "does this column exist" by running a `SELECT`
and reading the **error** as the answer — `no such column` *is* how `tableHasColumn`
says no. The retry read that same error as a stale memo, re-provisioned, ran the
migrations, which probed again. It presented as one test hanging and a package
`FAIL` with no failing test. Guarded twice: `tableHasColumn` bypasses the retry
entirely, and a context marker makes any statement issued *by* provisioning
ineligible for it.

### 2. `backlinks` was unbounded

And it was the only list-shaped op that read no `limit` at all, so a caller had no
way to ask for less. Inbound degree is the direction that concentrates — a hub
entity collects every fact written about it, and the maximum in-degree on the
304-chunk benchmark store is 162, growing with the corpus. Now bounded, and honours
`limit` like its siblings.

### 3. Tag writes were one statement per tag

The per-row `WHERE NOT EXISTS` guard is portable idempotence, not the correctness
barrier — `PRIMARY KEY (id, tag)` is — so one read plus one batched write gives the
same result in 2 round trips instead of N. The race is unchanged, not introduced:
two concurrent adders could always both pass the per-row guard and collide on the
key.

### 4. `set_asset` decoded before it checked the cap

The encoded length is an exact upper bound, so an oversize image no longer has to be
materialised in order to be rejected.

### What was tested (commit 2)

- `TestSchemaMemo_SelfHealsWhenAScopeGoesStale` — rewinds a scope *without* touching
  the memo, so the process still believes it is provisioned, and asserts an ordinary
  statement recovers on its own.
- `TestSchemaMemo_ProbesDoNotRecurse` — pins the guard. ⚠️ Without it this test does
  not fail, it **hangs**; that is the failure mode being guarded.
- `EnsureOnce`/`ForgetEnsured` validate their `ScopeKey`, which
  `TestScopeKey_EveryEntryPointValidates` caught me omitting: the memo id joins the
  three components with the reserved `0x1f` separator, so a component carrying that
  byte would alias two scopes and let one scope's provisioning mark another as done.
  Fixed rather than added to the exemption list.
- ⚠️ **One test changed, and the contract it tested genuinely moved.**
  `TestMigrateFactObservation_ReachesAnExistingScope` called `ensureSchema` twice in
  one process and expected the second to migrate. `ensureSchema` is once-per-process
  now, so the scenario it is really about — a **new process meeting an old scope** —
  is a dropped memo, and the test drops it before re-running. The guard is unchanged:
  does the migration reach a scope that already holds data? Worth a reviewer's eye:
  if `ensureSchema` should keep always-provisioning semantics, this is the line to
  push back on.
- `internal/sqlmem`, `internal/tools/builtin` and `internal/api/http` all pass.


denn-gubsky:
`-race` run at the full suite's own timeout, as promised in the description:

```
ok  github.com/denn-gubsky/loomcycle/internal/tools/builtin  781.464s
```

Green, **zero data races**.

⚠️ One thing I do *not* want read into that number: 781s here vs the 1475s baseline quoted in the description is **not** a like-for-like comparison and is not a speedup claim. The 1475s figure came from the package running inside the full parallel `./...` suite on a machine that was also hosting a benchmark server and Postgres; this run was the package alone on a quiet machine. The only controlled measurement in this PR is the test-level one — `TestGraphNeighbours_CapsRowsPerHop` at 95s versus 1.2s, same machine, same run, before and after making the cap a parameter.

The earlier `-timeout 900s` failure I mentioned while preparing this was my own error: 900s was never going to fit a package whose `-race` baseline is ~25 minutes. No defect, and not caused by the new tests.

PR #1279, opened 2026-09-18
feat(runs): promote a running agent to interactive, and let a run ask

denn-gubsky:
An agent could not be converted to interactive once it was going, so an operator watching one head the wrong way had no way to take hold of it and correct it. Two things were missing.

## `interactive` was captured at start

The park decision read a bool fixed when the loop began — so a run that started plain could never be told to wait, which is exactly what an operator wants *after* the run is in flight.

It is now read **at the boundary** through an optional callback: `nil` for every run that was never retuned (almost all of them), and otherwise a read of the run's stored configuration — correct across replicas for free, where an in-process flag would not be. One row read per turn boundary, against a turn that contains a model call.

The callback **fails to the start-time answer** — store error, undecodable record, no store at all — because a read that did not work must never flip a run between parking and finishing.

It works in **both directions**. A run started interactive can be released; a one-way door would be a worse bug than the one reported.

Three uses of the flag stay static on purpose: the unbounded-iteration lift, the auto context-mode choice, and `StartParked`. Promoting a run so someone can correct it should not also silently remove its iteration bound or rewrite how its history is kept — `max_iterations` is separately overridable.

The field is a **pointer** because three states differ: absent keeps what the run started as, `true` promotes, `false` releases.

## ⚠️ `Interruption` moved out of the "reach" class

This is a security-posture change, so it is stated rather than buried. It sat beside `MemoryScopes` and `Volumes` under *"what data and hosts the agent can touch"* — and it touches neither: an interruption **blocks and waits for a person**. The exposure is **liveness**, which is the class `unbounded_iterations` already occupies and which `run_timeout_seconds` and the interruption's own timeout bound. An untrusted trigger still cannot set it: webhook / A2A / scheduled runs build their input from the definition.

The argument lives in the classification table itself, not in a commit nobody re-reads.

## Every surface, driven by the guard

Adding the two names to `overrideWireNames` turned each unsatisfied surface red, and I fixed them until green — rather than surveying by hand. **The derived MCP check did it unprompted**: putting the fields on `connector.SpawnRunRequest` made `TestSpawnSchemas_AdvertiseEveryFieldTheyAccept` fail by itself, naming both tools and both fields. That is what deriving the expectation from the accepted shape is for.

Covered: both Go shapes, the proto (new `Interruption` message + fields on `RunRequest`/`ContinueRequest`), the gRPC arg struct and **both** construction sites, the MCP spawn schemas, the TS type + serializer, and the Python builder including `_run_request_from_dict` — the third enumeration that got forgotten last round.

`nil` in stays `nil` out at every mapping: "the caller said nothing" must reach the runner as *inherit the definition*, never as a zero-valued block that would **disable** interruptions the definition allows.

## Three things worth flagging

**A duplicate list removed in passing.** `applyOverridesToWire` declared its parameter as an inline structural type restating all twelve names — a second copy of the list inside the signature of the function whose only job is to be the single one. It now takes `RunOverrideOptions` itself.

**`isZero` was a struct comparison.** Adding a pointer-to-slice-bearing type broke `==`, and the tempting fix (`reflect.DeepEqual`) would have compiled and then read a zero-valued pointer as "supplied" — inverting the answer. Hand-written, with a test that a body naming only a new field is not empty.

**The suite caught a nil-store panic that review did not.** The callback runs at *every* turn boundary, and a server built without a store dereferenced nil on every run — surfacing as a truncated SSE stream missing its `done` frame, not an obvious crash. Guarded and pinned.

## What was tested

- `go test -race -timeout 1800s ./...` — exit 0, 103 packages, no FAIL/race. TS **308**, Python **147**.
- The boundary decision prefers the live answer in both directions.
- **The crossing** — retune → store → the loop's own decision — probed by dropping the merge: it reports that the promotion reached the store and not the decision.
- An unreadable run keeps what it started as; promoting leaves the rest of the run's configuration alone; interruption enables for one run.
- Parity guard green over every surface with the new names; derived MCP check green; **stub gate clean after regeneration** (both sets, verified locally).

Independent of #1277 / #1278 — it extends the shared wire struct, so the `/retune` route and the gRPC RPCs pick these fields up automatically when those land.


denn-gubsky:
**Rebased onto `main` (now carrying #1277 and #1278).** `MERGEABLE`.

This was not a mechanical conflict fix — those two PRs landed surfaces these fields had to reach, so the merge needed integration rather than resolution:

- **`RetuneRunRequest` and `RunInputRequest`** now exist and carried neither field; both do now, so `interactive`/`interruption` work on the dedicated `/retune` route and the gRPC RPC.
- **`connector.RunOverrides`** (the retune shape, new in #1278) gained both, and the **HTTP connector impl** carries them through to `retuneRun`.

Without that, the merge would have produced two features that each pass their own tests and do not compose.

**Two things the merge forced, both worth keeping:**

`connector.RunOverrides.IsZero` was `o == (RunOverrides{})`. Adding a pointer-to-slice-bearing type breaks that comparison, and the tempting fix — `reflect.DeepEqual` — compiles and then reads a zero-valued pointer as "supplied", inverting the answer. Hand-written, matching its sibling on the wire struct.

`overridesFromProto` had reached **twelve positional parameters**, six of them `*bool`/`*int32`. Two more would have made a transposed pair compile silently. It now takes a keyed struct.

**The tests got better because of the merge.** They were written against `/input`'s `overrides` because `/retune` did not exist on this branch. They now use `/retune` and assert the provider sees **no additional request** — which is the actual reported scenario: take hold of a running agent without putting a message in its transcript that nobody wanted to send.

Also worth recording: git's merge silently swallowed three closing braces while resolving. `gofmt`/`go vet` caught each — a resolution that looks clean is not one.

**Verified after the rebase:** `go test -race -timeout 1800s ./...` exit 0, 103 packages, no FAIL/race. TS **313**, Python **150**. Both parity guards green with the new names in the canonical list, the derived MCP schema check green, and the **stub gate clean** — regenerating from the merged proto yields zero drift in Go and Python.

PR #1298, opened 2026-09-18
feat(memory): runtime-supplied retrieval — +31pp on a local answerer (gate NOT met; see retraction)

denn-gubsky:
## Why

The local-model plan rested on a premise L1 falsified: the gap was never reading. On
plain reading the local models sit within 3pp of the cloud model, and ornith's
ceiling (0.8087) sat **47 points above its own store arm**. It was starved, not weak.

This ships the two levers that feed it, and measures both.

## What

**L2 — `recall_attach_traces`.** `recall_include_turns` attaches the turn each
recalled *fact* came from; that is fact-anchored and can only reach turns some fact
was already extracted from. This also searches the trace index with the caller's own
query and returns those turns as `source_turns`, beside `memories` — a separate
block with its own budget, because fusing them is what `ErrTracesNotCombinable`
refuses. **Deliberately no tool parameter**, with a test asserting its absence: the
sibling shipped with one and it was passed on 51 of 128 calls.

**L3 — `{{memory:recalled_context}}`.** Renders FACTS and TURNS blocks into the
system prompt before the first token, so the reader can hold no tools. A sibling of
`search_request` rather than a change to it. Pinned as *not* trusted — it carries raw
user turns, so it renders inside the data frame.

## Measured — five arms, one host, same 150 questions and judge

| arm | accuracy | strict | abstention | p50 | calls | gate |
|---|---|---|---|---|---|---|
| control | 0.3758 | 0.2953 | 0.387 | 6.8s | 151 | fail |
| P1 fact-anchored | 0.4799 | 0.3758 | 0.287 | 7.7s | 150 | fail |
| L3 tool-free | 0.5933 | 0.4933 | 0.187 | **1.7s** | **0** | fail |
| **L2 runtime traces** | **0.6858** | **0.6014** | 0.113 | 8.3s | 150 | **PASS** |
| oracle (ceiling) | 0.8048 | 0.7397 | 0.087 | 1.0s | 0 | — |

**First local answerer to clear the gate** (accuracy ≥ 0.67 and strict ≥ 0.60). L2
captures 72.3% of the headroom to the ceiling and is strictly dominant over the
control: **+54/−0, p=1.1e-16**. Every rung is significant — control→P1 +24/−3
(4.9e-05), P1→L3 +37/−17 (9.1e-03), L3→L2 +29/−12 (1.2e-02), P1→L2 +43/−5 (1.4e-08).

Nothing changes the model or the prompt — only what the runtime hands over.

**Temporal was never a temporal problem:** 0.1892 → 0.7000, won 21 / lost 0. The
answer was handing over the turn, which carries its own timestamp while the distilled
fact is tenseless. L4 should be re-costed before it is built.

**P1's grant replicates on a second model:** qwen3.6 +11.4pp, ornith +10.4pp.

## ⚠️ The tool-free reader loses, and that settles what L1 could not

L1 changed evidence and protocol together and could only bound the combination — I
flagged that at the time. Removing the protocol while the runtime still supplies the
evidence costs **9.3pp (p=0.0115)**, so the tool loop has *positive* value: L2's query
is model-composed, it carries both kinds of turn, and it can issue a second recall
when the first is thin. The coerced-arm collapse to 0.0034 was about **mandating a
procedure**, not about holding a tool.

L3 remains the cost story: 50.7% of the headroom at 20% of the latency, zero tool
calls, on any model that can read a dated block.

## Tested

- Six tests for L2, four for L3; full suite green, 103 packages, `go vet` clean.
- The carry guard derives its expectation from the sibling's count rather than a
  hand-written list, and was verified to **fail** when one carry site is removed.
- Three drift guards fired and were satisfied properly rather than suppressed: the TS
  mirror (the register may only shrink, so the field was added), the overridability
  classification (`notOverridable` — it is read reach, wider than its sibling), and
  the lookup field census.
- Controls verified, not assumed: the control elected traces on **0 of 151** calls;
  trace-index writes disabled after a smoke run indexed its own question; the reader
  prompt asserted byte-identical across control/P1/L2; peak context 14112/32768.

Results, summaries and method notes: `bench/results/locomo/2026-09-19-l2-l3-spark/`.


denn-gubsky:
## ⚠️ Retraction: the gate claim in the PR description does not hold

A replicate of the L2 arm (identical definition, same host and store) was run
specifically because draw 1's strict sat **0.0014** above the 0.60 floor — 89 correct
of 148, where one question flips it.

| draw | accuracy | strict | correct | gate |
|---|---|---|---|---|
| L2 draw 1 | 0.6858 | 0.6014 | 89/148 | PASS |
| **L2 draw 2** | **0.6588** | **0.5878** | **87/148** | **FAIL** |
| **mean** | **0.6723** (+0.0023) | **0.5946** (−0.0054) | — | **FAIL** |

**"First local answerer to clear the gate" is withdrawn.** Accuracy clears on the
mean; strict misses. The gate needs both. It also asks for two conversations, which
no local arm has attempted — so even a passing strict was one of two conditions.

## What is unaffected

The lever is far outside the churn that swallowed the threshold:

| pair | draw 1 | draw 2 |
|---|---|---|
| control → L2 | +54/−0, p=1.1e-16 | +51/−4, p=2.1e-11 |

L2 still moves ornith 0.3758 → 0.66–0.69, captures 66–72% of the headroom to its
reading ceiling, and cuts abstention 0.387 → 0.11–0.15. The L2-vs-L3 result (the
tool loop is worth 9.3pp), the temporal result (0.1892 → 0.7000, won 21 / lost 0),
and the host-transfer decomposition are all unchanged. **No code changes.**

The honest headline is: **L2 is worth ~+31pp on a local answerer and lands on the
strict criterion rather than over it.**

The lesson, which this repo already had written down: a threshold claim needs a
different standard of evidence from an effect claim. +31pp is safe on one draw; a
0.0014 margin against a fixed floor is one question. Report a threshold as a mean
over draws with its spread, or do not report it as met.

Research document updated — §7b retitled and §7b-i carries the retraction.

denn-gubsky:
## ⚠️ A coverage gap in this PR's grant, found by the second conversation

conv-30 (the gate's second conversation) shows **L2 at +12.3pp against conv-26's
+29.6pp**, and the cause is a gap in `recall_attach_traces` rather than a property of
the corpus.

**The grant enriches `op=recall` only, and the model elects the op:**

| corpus | `op=recall` | `op=search` | L2 fires on |
|---|---|---|---|
| conv-26 | 136 | 14 | 91% |
| conv-30 | 36 | 49 | **42%** |

Per question the partition is total, and the inert half is its own control:

| conv-30 questions | n | got turns | control | L2 | Δ |
|---|---|---|---|---|---|
| used `recall` — L2 fires | 36 | 36 (100%) | 0.5139 | **0.7917** | **+27.8pp** |
| used `search` — L2 inert | 45 | 0 (0%) | 0.3778 | 0.3778 | **+0.0pp** |

The inert half moves by **exactly zero**, which rules out reverse causation. And
+27.8pp on the fired half matches conv-26's coverage-normalised +32.7pp. **L2 is
worth ~+28pp wherever it fires; the whole shortfall is coverage.**

**This is the invocation problem one level down.** The grant removed the model's
decision about *whether to search traces* but left its decision about *which op to
call*, and `search` bypasses it. A tool parameter is a decision the model makes
(51/128) — and so is the op.

**Suggested fix before merge:** make the grant op-agnostic (apply it to `search` as
well as `recall`), or rename it for what it governs rather than for one op. I have
not made the change — it alters a shipped surface and the maintainer should choose
the shape. But every L2 number in the programme is currently a coverage-weighted
average of ~+28pp and an op distribution nobody controlled.

Also: **the gate fails on conv-30** by 11 and 12 points (0.5617 / 0.4815 against
0.67 / 0.60), so the retraction in the earlier comment stands and hardens — ~0.67
was a conv-26 property.

Results: `bench/results/locomo/2026-09-19-conv30/`. Research doc: §7b-ii.

PR #1304, opened 2026-09-19
fix(memory): the trace grant covers every op — and the gate is met on both conversations

denn-gubsky:
Follow-up to #1298. Two gaps the **second LoCoMo conversation** exposed, plus the
conv-30 results that found them.

## 1. The grant enriched `op=recall` only — and the op is the model's choice

`recall_attach_traces` shipped attached to one op. Same agent definition, two
conversations:

| corpus | `op=recall` | `op=search` | grant fires on |
|---|---|---|---|
| conv-26 | 136 | 14 | **91%** |
| conv-30 | 36 | 49 | **42%** |

Per question the partition is total, and the inert half is its own control:

| conv-30 questions | n | got turns | control | L2 | Δ |
|---|---|---|---|---|---|
| used `recall` — fires | 36 | 36 (100%) | 0.5139 | **0.7917** | **+27.8pp** |
| used `search` — inert | 45 | 0 (0%) | 0.3778 | 0.3778 | **+0.0pp** |

The inert half moves by **exactly zero**, which rules out reverse causation — if
`search` merely marked harder questions, the arm would still have shifted them. And
+27.8pp on the fired half matches conv-26's coverage-normalised +32.7pp. **The lever
is worth ~+28pp wherever it fires; the whole shortfall was coverage.**

This is the tool-parameter lesson one level down. A caller-elected parameter was
passed on 51 of 128 calls, so the decision moved to an operator grant — and then the
**op** quietly became the new decision. An operator grant has to hold wherever the
capability is reachable, not on the one path it was first written for.

⚠️ Suppressed on a **traces-only** search: those turns arrive as the result's
entries, and appending the same rows again under `source_turns` is duplication the
reader must reconcile. `ErrTracesNotCombinable` already refuses mixing traces with
other sources, so "asked for traces" is exactly a traces-only query.

## 2. The trace index stored `time.Now()` — when it was INDEXED, not when it was said

Nothing carried the turn's own instant, so a trace search could answer *what was
said* and never *when*. A backfill run months after a conversation stamped every turn
in it with the same afternoon.

That is the half the temporal questions need: a distilled fact is tenseless and the
**turn** is what carries the date — which is why attaching turns moved the temporal
slice at all. **LoCoMo hid this**, because its harness pre-stamps turn text as
`[7:18 pm on 27 May, 2023]`; a live-indexed production turn had no date whatsoever.

`conversationTurn` now carries `At` from the transcript event, and the two **replay**
paths (backfill, and assistant turns at `finishRun`) take the turn's own stamp. The
live user path keeps `now()` **with a comment saying why** — it indexes a turn as it
arrives, so `now()` *is* the instant. An assistant turn is stamped when it **began**,
matching the user turn beside it, rather than when its last delta landed.

This also unblocks deterministic temporal grounding, which cannot render a weekday or
a day-offset for a turn whose instant was never stored.

## 3. conv-30 results (the gate's second conversation)

| arm | accuracy | strict | gate |
|---|---|---|---|
| control | 0.4383 | 0.3827 | fail |
| L2 | 0.5617 | 0.4815 | **fail** |
| oracle (ceiling) | 0.7785 | 0.7342 | — |

**The gate fails by 11 and 12 points** — so ~0.67 was a conv-26 property, and the
retraction on #1298 hardens. Built in an isolated store so conv-26's arms survived.

Also recorded: conv-30's temporal questions are *easy* for a cloud reader (deepseek
0.9423) and hard for ornith **holding the gold turns** (0.6154) — a 33-point gap
where retrieval cannot be blamed.

## Tested

- 8 tests in `memory_recall_traces_test.go`, 4 in `history_turn_instant_test.go`.
- Both new structural guards **verified to fail when their change is reverted**:
  removing the search-path attachment, and stubbing `turn.At` in the backfill.
- Full suite green on current `main`, `go vet` and `gofmt` clean.

Branched fresh off `main` rather than reusing #1298's merged branch, so none of
#1300–#1302 is touched.

⚠️ **Existing trace rows keep their old timestamps until re-backfilled.**


denn-gubsky:
## ⭐ The fix clears the gate — both conversations, replicated

Re-ran both corpora with this PR's binary. Arm definitions byte-identical to the
pre-fix runs; **only the binary differs.**

| arm | accuracy | strict | gate |
|---|---|---|---|
| conv-26 draw 1 | 0.7013 | 0.6242 | PASS |
| conv-26 draw 2 | 0.6939 | 0.6190 | PASS |
| **conv-26 mean** | **0.6976** | **0.6216** | **PASS** |
| **conv-30** | **0.7654** | **0.7160** | **PASS** |
| **both-conversation mean** | **0.7315** | **0.6688** | **PASS** |

Against control: **+57/−1, p=4.1e-16** (conv-26), **+32/−1, p=7.9e-09** (conv-30).

**This is not the claim retracted on #1298**, and that retraction still stands for the
implementation it described — the code changed under it. That one rested on **+0.34 of
a question** from a single draw and died on its replicate. This clears on every
individual draw, on the mean, and on both conversations, with margins of 4.1/3.2
questions (conv-26) and 7.7/9.4 (conv-30). The two conv-26 draws differ by 0.0075.

**The gains track the gaps they closed**, which is what makes the coverage explanation
credible rather than fitted:

| corpus | coverage before | before | after | Δ |
|---|---|---|---|---|
| conv-26 | 91% of calls | 0.6723 | 0.6976 | +2.5pp |
| conv-30 | 42% of calls | 0.5617 | 0.7654 | **+20.4pp** |

A post-hoc story would not have predicted the *small* conv-26 number. conv-30's
questions receiving turns went 36/81 → **81/81** with the op distribution unchanged
(`search` 47, `recall` 36) — the model behaves identically, the runtime now serves it.
It captures 96.2% of its headroom and sits at 98.3% of its reading ceiling.

## ⚠️ One conclusion this overturns — relevant to reviewers

**Temporal is not a local-reader reading limit**, which I claimed from the oracle arm
(ornith 0.6154 holding gold evidence vs deepseek 0.9423). The fixed conv-30 arm reaches
**0.9038 on temporal — beating the oracle.** That is possible because the oracle renders
only LoCoMo's *annotated* evidence, and that annotation is off by one; real retrieval
finds dated turns it omits.

So an oracle arm on this corpus is a ceiling for **reading the annotation**, not for
reading what is retrievable. The deterministic temporal-grounding lever should **not**
be built on the evidence that motivated it — the deficit was coverage.

The trace-instant capture in this PR still stands on its own merits: a trace search
could not answer *when* for any live-indexed turn.

Results committed: `bench/results/locomo/2026-09-19-gate-met/`. Research doc: §7b-iii.

denn-gubsky:
## `recallTraceTopK` 6 → 24 — measured, and it lifts both conversations

One more commit on this branch. `attachQuestionTurns` had two bounds and only one
ever engaged: **300 of 300 recalls returned exactly 6 turns**, while the block
averaged **1271 chars against a 6000-char budget** — 4× headroom the budget never
used.

| | k=6 (mean of 2) | k=24 (mean of 2) | Δ |
|---|---|---|---|
| conv-26 | 0.6976 / 0.6216 | **0.7457 / 0.6826** | +4.8pp |
| conv-30 | 0.7593 / 0.7160 | **0.8013 / 0.7454** | +4.2pp |
| **both** | 0.7284 / 0.6688 | **0.7735 / 0.7140** | |

Magnitudes agree within 0.6pp across corpora with different question mixes and turn
lengths; within-condition spreads 0.0018–0.0123.

**The claim rests on the magnitudes, not the p-values.** conv-26 reaches p=0.017–0.041
on three of four cross-pairings; **conv-30 reaches p=0.18–0.73 on none** — 81 questions
and a high baseline leave only 5–7 moving. conv-30 is consistent-and-underpowered, not
independent confirmation.

**Why not higher:** k=48 is a plateau — +11/−11 (p=1.0) and +11/−13 (p=0.839) for
+1.1s latency. At 48 the budget finally binds (1 of 150 recalls reached the count) and
buys nothing. 24 is the smaller of two values that perform identically.

**It is not free:** k=24 loses 5–6 questions k=6 answered correctly. A long-turn
corpus would pay more of that dilution cost and collect less — the budget is what
protects that case, which is the argument for letting it be the real bound.

## ⚠️ A metric in my earlier comments is retired

I quoted figures like *"captures 96.2% of headroom"* and *"92.5% of the ceiling"*
against the oracle arm. **conv-30 at k=24 scores 0.8013 against an oracle of 0.7785 —
108.2% of it.** The oracle renders only LoCoMo's *annotated* evidence, which is off by
one, so it bounds *reading the annotation*, not achievable accuracy. Those percentages
should be disregarded; the tables now report control→arm deltas and McNemar only.

Results: `bench/results/locomo/2026-09-20-topk/`.

PR #1318, opened 2026-09-21
fix(context): the effective report answers from the run, and a decline says how full

denn-gubsky:
Three gaps reported from the panel built against this surface. Each one is a value that exists and never arrives.

## 1. `inert` was computed against the wrong definition

`GET /v1/runs/{id}/effective-config` merges the run's overrides onto the stored definition and hands the result to `InertContextSettings` — but it passed `Routing`, `Resources` and `Tuning` and **not** `Context`.

So the advisory array answered from the stored definition while `fields.context` **in the same response** answered from the run: one payload, two verdicts about one setting. Seeing a per-run override introduce — or clear — a trap on a definition that is not that way by itself is the entire reason this array exists over the boot-time warning, and it structurally could not.

`runOverrides` gains the field and `effectiveDef` merges it **per field**, matching how the run path applies it: a run that sets only `mode` must not blank the agent's `keep_last_n`.

## 2. `used_tokens` was declared on every decline and set by nothing

Since `ContextDistillDeclinedInfo` existed, a decline reported the window size and left the reader to guess how full it was — the one number that decides whether a decline is urgent.

That is worse than an absent field: a consumer sees `0` and reads *"0 tokens in use"*, not *"not reported"*. `maybeRecap`, `maybeAutoCompact` and the shared summariser-failure builder now take the footprint and stamp it, at all six construction sites.

## 3. The TS client could not see `inert` at all

`EffectiveConfigResponse` declared `fields` and stopped, so the advisory half of the report was unreachable from a typed consumer — and it is the half `fields` structurally cannot express: a setting can be **in force and inert**, because a different setting disables it. Adds `inert` and the `InertContextSetting` shape.

## Tested

Both new Go tests fail on the unfixed code.

- `TestEffectiveConfig_InertIsComputedAgainstTheRunsOwnContext` — asserted through the **endpoint**, not on `effectiveDef`, because the omission was at the call site: the merge helper was always capable of this and was simply never handed the field. Non-vacuity first: the definition alone must report the trap, or the second half proves nothing.
- `TestDistillDeclines_EveryVerdictCarriesTheFootprint` — swept by **struct literal** across both distillers and the shared builder, not by exercise: a reason reachable only under a configuration the test does not build would be missed, and the literals are the complete set by construction. Fatals below five found, so it cannot quietly read nothing. Probed: removing the stamping flags all six.
- `TestDistillDeclines_TheFootprintReachesTheEvent` — the behavioural half, so the value has to arrive on the wire and not merely appear in the source.

`tsc --noEmit` clean. `internal/loop`, `internal/api/http`, `internal/config` green.

## Merge order

Touches `internal/loop/loop.go` and `distill_declined_test.go` in regions disjoint from #1316, but a rebase may still be wanted depending on which lands first — happy to take it in either order.


denn-gubsky:
Rebased onto `main` now that #1316 and #1317 have landed.

One conflict, in `internal/loop/distill_declined_test.go`, and it was a both-added-at-EOF: #1316 appended `TestRun_NoContextReportBeforeTheFirstMeasuredTurn` and this branch appended the two `TestDistillDeclines_*`. Both kept, in that order. `loop.go` and `adapters/ts/src/types.ts` auto-merged — verified by hand that both halves survived in each: `footprintMeasured` still gates the reports **and** all six decline literals still stamp `UsedTokens`; the union carries `context_exhausted` **and** `EffectiveConfigResponse` carries `inert`.

Re-verified after the rebase: `internal/loop`, `internal/api/http`, `internal/config`, `internal/providers` green, `tsc --noEmit` clean, full `-race` suite running.

PR #1321, opened 2026-09-21
feat(providers): the emit_state contract is on the wire, not only in the prompt

denn-gubsky:
RFC DG P1 — `/loomcycle/rfcs/forced-tool-choice`.

## Why

`grep -rn "tool_choice" internal/` returned **one** hit before this change: a comment listing OpenAI-compat fields the gateway ignores. loomcycle has never sent `tool_choice` on any driver, so the stateful loop's "call `emit_state` exactly once" lived entirely in the system prompt. The model that failed on `ornith-1.5:35b` was not disobeying an instruction the wire ever sent — it was asked politely, in English, and it declined.

## What

`providers.ToolChoice{Mode, Name}` on `Request`, a `SupportsToolChoice` capability, and the mapping in each of the **four** wire encoders. `deepseek`, `vllm` and `llamacpp` inherit through `inner.Capabilities()` exactly as they do for vision, so this is four places, not ten.

The zero value is auto, so **every unopted request is byte-identical on every driver**.

The three dialects genuinely differ, which is why no assertion is shared between them:

| Driver | none | any tool | one named tool |
|---|---|---|---|
| anthropic | `{"type":"none"}` | `{"type":"any"}` | `{"type":"tool","name":X}` |
| openai (+3) | `"none"` | `"required"` | `{"type":"function","function":{"name":X}}` |
| gemini | `mode:NONE` | `mode:ANY` | `mode:ANY` + `allowedFunctionNames:[X]` |

⚠️ **Gemini has no "call exactly this one" mode** — a single forced tool is `ANY` narrowed by a one-element allow-list.

⚠️ **OpenAI's named form is the nested one.** The flat `{"type":"function","name":X}` is the Responses API's; this driver posts to `/chat/completions`. Some gateways accept the wrong shape as a silent no-op — forcing that does nothing, reported as success.

## Two judgement calls worth reviewing

**Anthropic rejects a forced choice under extended thinking**, and the pair is reachable from ordinary configuration rather than a mistake: this driver attaches `thinking` whenever an agent declares an `effort` hint on a reasoning-capable model. Thinking wins and the choice is dropped with a log line — same precedence the driver already applies to temperature, and for the same reason: the operator opted into reasoning explicitly, while forcing is an optimisation of a contract the prompt still carries. `none` survives, since the API accepts it there and an over-broad guard would silently re-enable tools on a run that asked for none.

**Ollama is non-mandatory, by your decision.** `/api/chat` has no `tool_choice` and neither does its OpenAI shim, so `SupportsToolChoice` is an explicit `false` and a forced request **degrades rather than being refused**. Refusing would make every local-model agent unrunnable to buy a guarantee it never had. The one lever Ollama does have — narrowing `tools` to the single tool you want — was already fully pulled by the stateful loop when the incident happened; constrained decoding via `format` is the real answer there and is P2.

## Tested

- **`TestToolChoice_EveryDriverThatClaimsItSendsIt`** — ⚠️ derived from the **capability**, not from a list of drivers, because a list is a second copy of the claim and the copy is what drifts. Every registered driver is built against a capture server, asked what it supports, and the ones saying yes must change the body. Asserted through `Call()` rather than each package's private builder: a driver could map `ToolChoice` in a helper nothing calls and still pass a unit test of the helper. **Probed** by deleting one mapping line — it failed *four* drivers at once, which is the delegation fan-out working.
- **`TestToolChoice_ADriverWithoutItStillAcceptsTheRequest`** — the other half, encoding the decision above: ollama must not refuse.
- Per-driver wire-shape tests for all three dialects + an auto-sends-nothing test each. Deliberately not shared: a helper checking "the body mentions the tool" would pass on a driver emitting another's dialect.
- **`TestRun_Stateful_AsksForTheEmitStateToolOnTheWire`** — the crossing. A mapping nothing sets is as inert as no mapping, and that gap is invisible from either side.
- ⚠️ The Anthropic thinking test needed `max_tokens` raised: the driver skips thinking when the 8192-token high budget can't fit under it, so the first draft asserted nothing. The non-vacuity guard caught it.

## Not in this change

- **P3, the operator-facing per-agent knob** — it needs `until` semantics first: a choice pinned for a whole run can never let the model answer, so the run hits `max_iterations`.
- **P2, structured output** for the providers that cannot force (`format` on ollama-local, `response_format`, `responseSchema`).
- **The fallback re-check.** Deliberate: with Ollama degrading rather than refusing there is nothing to suppress, and RFC AT's refuse-the-swap rule only becomes necessary once forcing is mandatory.
- The "this provider could not be forced" note in the stateful terminal error — that text belongs to #1320 and would conflict; it composes naturally once both land.


denn-gubsky:
Rebased onto `main` now that #1320 and #1322 have landed. One conflict, in `internal/loop/stateful_test.go` — both-added-at-EOF again (#1320's tests vs the tool-choice crossing test); both kept. `stateful.go` auto-merged and I checked by hand that both halves survived: #1320's `errNoEmitState`/retry machinery **and** the `ToolChoice` on the request.

Also added the piece this PR had deferred, since its terminal error now exists on main: **the error says when the tool could not be forced.** Without it, "the model ignored the constraint" and "there was no constraint" read identically, and they call for opposite next moves — replace the model, or move the agent to a provider that has a `tool_choice` at all. Tested both directions in one table; asserting only the first would pass on a note that fires unconditionally.

`internal/loop`, `internal/providers/...`, `internal/api/http` green.

PR #1323, opened 2026-09-21
feat(context): recap can run on its own model, and a guard finds the next gap

denn-gubsky:
`compaction:` has had a `model:` override since compaction shipped — run the summary call on a cheaper model from the same provider. `context:` never got the same knob, so a **recap always ran on the agent's own model**.

## ⚠️ The point is the model's behaviour, not its context window

The report that prompted this suggested a summarizer with a 1.5–2× bigger window. A recap reads a span of transcript and writes a short note, so it is **never** the call that runs out of window. What breaks it is a *reasoning* model, which spends its output budget thinking before it writes. What `context.model` buys is a **cheap non-thinking summarizer beside a thinking chat model**.

```yaml
context:
  mode: recap
  model: qwen3.6-instruct     # non-thinking; the chat model stays qwen3.6
```

## Carried across every surface

config (field + `IsZero` + `Clone` + `MergeContext`) · the agent signature · the AgentDef overlay and its read adapter · the loop · proto + gRPC mapper · the MCP per-run schema · both adapters · the editor's field registry.

## And a derived guard

That list is six places, and this project has shipped the partial version of it repeatedly: the MCP schema that hid the per-run overrides, the Python stubs that dropped them, sixteen event types absent from the TS union.

`TestContextFields_ReachEverySurface` reflects over `config.Context`'s json tags and requires each one on each surface. A hand-written field list would be a second copy of the struct; **the struct is the list.**

**It found a real pre-existing gap on its first run:** `recall` and `harvest_to_memory` were missing from the MCP per-run context schema, so an MCP caller could not set either. Both added here.

## ⚠️ Probing the guard is what made it worth having

The first version searched each file whole and **passed with the new field deleted**. Both nested surfaces carry an unrelated `model` — the agent's own routing model in the field registry, four other schemas in the MCP file — and a one-word tag like `recall` or `mode` occurs in ordinary prose in every one of these files.

It now narrows each file to the region describing `context` by balancing brackets from an anchor, and **fails when the anchor is gone** rather than silently reading nothing. Re-probed once per surface after the fix: all six caught.

## Tested

- `TestRecap_UsesTheConfiguredSummarizerModel` — and the complement, that unset still uses the run's own model, because an agent that never configured this must not change behaviour.
- `TestContextModel_ChangesTheAgentSignature` — `IsZero`, `Clone` and `MergeContext` all carry it. A content-identifying field omitted from the signature would let a fork change how an agent behaves while claiming to be byte-identical.
- `go test ./internal/...` clean; `tsc --noEmit` + the TS context-options suite green.

## Notes for review

- ⚠️ The gRPC stub regeneration is committed for `loomcycle.pb.go` **only**. `loomcycle_grpc.pb.go` changed by comment *reflow* alone under my protoc (7.36.2 vs CI's 7.35.1) with no schema difference, and CI compares those files byte for byte apart from version stamps — committing the reflow would fail the very gate that exists to catch real drift.
- The test fixture duplicates #1322's `longChat` under another name so the two branches don't collide in package `loop`; collapse them on whichever rebases second.


denn-gubsky:
Rebased onto `main` (clean) and **fixed the red CI**.

**The failure was mine, in `@loomcycle/def-fields`:** I wrote `type: "string"` on the new control, and `FieldType` has no such variant — its text variant is `"text"`, as the sibling model control three groups up already uses.

⚠️ Worth recording *why I missed it*: I ran the package's **tests** locally and they pass either way, because the registry's drift guard checks which **keys** are present, not that each one type-checks. CI runs `npm run typecheck && npm test`; I ran only the second half.

Also collapsed the duplicated test fixture onto #1322's `longChat` now that it has landed — the two branches each needed a conversation the default `keep_last_n: 6` can still split and couldn't share one while both were open.

Re-verified after the rebase: `npm run typecheck` + `npm test` in def-fields, the context-surface parity guard, and `go test ./internal/...` all clean.

